PR #4, opened 2020-02-12
Add basic WebP format support.

ImagicTheCat:
As discussed [here](https://love2d.org/forums/viewtopic.php?f=4&t=88255).

It's a basic encoding (lossless) / decoding implementation. Tested under GNU/Linux. 
You're free to use it or not, but at least the work is done if the feature is desired at some point.


src/modules/image/magpie/WEBPHandler.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+	return encodedFormat == ENCODED_WEBP && rawFormat == PIXELFORMAT_RGBA8;
+}
+
+WEBPHandler::DecodedImage WEBPHandler::decode(Data *data)
+{
+	int width, height;
+	uint8_t *pdata = WebPDecodeRGBA((uint8_t*)data->getData(), data->getSize(), &width, &height); 
+
+	if(pdata){
```
slime73: As per our [code style guide](https://love2d.org/wiki/Code_Style), braces go on a new line and there should be a space between the `if` and the `(`.

slime73:
Is there a corresponding change to megasource (cmake dependencies) and the love-android repository along with this? Is libwebp trivial to build for macOS and iOS as well?

What sort of use case do you have in mind for webp files that isn't covered by png and jpeg? Adding and then maintaining a dependency like this isn't free for us, so its cost needs to be offset by solid use cases.

ImagicTheCat:
No I didn't take care of the whole cross-platform build process, this is only the implementation and a few changes to test on POSIX systems. I don't know how exactly LÖVE is maintained.

https://developers.google.com/speed/webp/download
Yes, libwebp has probably been designed to be compiled for all those targets. Also, the licence should allow for any kind of distribution in LÖVE.

I understand about the dependency issue, but WebP has potential. I showed some examples on the topic, the basic implementation would give:
- better compression than PNG and JPEG
- flexibility with lossy/lossless RGBA (drastically better compression for complex pictures with alpha, animation atlas, etc)

Then you could use the animation feature for interesting stuff, but that would require new APIs.

I don't know if it's worth the effort from your point of view. To me moving forward is important, so if I can get rid of PNG, JPEG and GIF at once in my work with improvements, I will. This pull request is mostly about showing how simple the libwebp API is.

MikuAuahDark:
> 
> 
> Is there a corresponding change to megasource (cmake dependencies) and the love-android repository along with this? Is libwebp trivial to build for macOS and iOS as well?
> 

I'm not sure about iOS and macOS, but I think libwebp integrates with Megasource since it has CMake scripts. It also looks like integrates well with love-android as it provides their own Android.mk file.

ImagicTheCat:
With hindsight, altough interesting, supporting WebP is not essential. There is work on new image formats; particularly JPEG XL looks promising.

It seems that the rate of progress is faster than the rate of adoption, maybe we need to hit a wall to reach a consensus.

slime73:
What I'm really hoping for with WebP is for someone to make a minimal single-file or very small decoding library for it, similar to stb_image or LodePNG, since the reference/official library is so big.

I never looked at the spec or anything though, so I suppose it's possible the format is complicated enough that something as small as LodePNG isn't ever going to happen for WebP.

PR #5, opened 2020-02-16
Replace Mpg123 with dr_mp3

MikuAuahDark:
See https://bitbucket.org/rude/love/issues/1499#comment-52786180

> Switching to one of those is probably a good idea, but I’ll probably hold off until love 12 since there’s a chance they aren’t able to correctly play back every possible mp3 file that mpg123 can. I think we had that problem for a little while when we switched png decoders.

Note that this pull request is incomplete:
* The macOS & iOS xcode project file must be updated to reflect the new file, but since I don't have mac/book, I'll left this to slime.
* The unix automagic stuff is not extensively tested. I just do search-and-replace to it.
* ~~I haven't fully tested if this really works (need to sleep as of writing). I'll update this once I'm able to test it in Windows and Linux.~~ I've tested this in Windows but the file loading performance is 4x slower than Mpg123. Not sure if this is worth it. It's a tradeoff between maintaining Mpg123 library vs. single-source MP3 decoder.


extra/nsis/love.nsi
```diff
@@ -16,8 +16,8 @@ InstallDirRegKey HKCU "Software\LOVE" ""
 !define MUI_WELCOMEFINISHPAGE_BITMAP "${LOVEICODIR}\left.bmp"
 !define MUI_UNWELCOMEFINISHPAGE_BITMAP "${LOVEICODIR}\left.bmp"
 
-!define MUI_WELCOMEPAGE_TITLE "L�VE Setup"
-!define MUI_WELCOMEPAGE_TEXT "This will install L�VE, the unquestionably awesome Lua game framework."
+!define MUI_WELCOMEPAGE_TITLE "L�VE Setup"
+!define MUI_WELCOMEPAGE_TEXT "This will install L�VE, the unquestionably awesome Lua game framework."
```
MikuAuahDark: Why this happends. I'll fix this when I wake up.

PR #1553, opened 2020-03-01
Changed to set the window title during creation rather than after

### LINKED ISSUE #1552: Window title is set after, not during, window creation.
Not really a major issue, but on Linux you can use some window managers to arrange applications when they start. e.g. Put them in specific locations, desktops, etc.

So for example, the WM can match an application's window title and then move it to a second desktop.

Currently with love, the window gets created and **then** it sets the window title.  This means that the WM thinks the application's window title is blank when it starts.

namtabmai:
Fixes #1552

Not really familiar with the LUA C api, but this appears to work for me. Have not tested on anything other than Linux.


src/modules/window/wrap_Window.cpp
```diff
@@ -94,6 +94,12 @@ static int readWindowSettings(lua_State *L, int idx, WindowSettings &settings)
 		settings.y = (int) luaL_optinteger(L, -1, 0);
 	}
 	lua_pop(L, 2);
+	lua_getfield(L, idx, settingName(Window::SETTING_TITLE));
+	if (!lua_isnoneornil(L, -1))
+	{
+		settings.title = luaL_checkstring(L, -1);
+	}
```
MikuAuahDark: As per our [code style](https://love2d.org/wiki/Code_Style), braces should be omitted whenever possible.
```suggestion
		settings.title = luaL_checkstring(L, -1);
```

src/scripts/boot.lua
```diff
@@ -550,8 +550,8 @@ function love.init()
 			usedpiscale = c.window.usedpiscale,
 			x = c.window.x,
 			y = c.window.y,
+			title = c.window.title or c.title
```
MikuAuahDark: I would use `title = c.window.title or "Untitled"` here, to be consistent with previous versions.

MikuAuahDark:
I most likely miss some things here, so I'll also let slime to review it. I'll test it in Linux when I have time.

src/scripts/boot.lua
namtabmai: I was simply moving the line 554 to here, but looking the documentation there doesn't appear to be a config option for an application title, only the window. Will change as suggested.

src/scripts/boot.lua
```diff
@@ -550,8 +550,8 @@ function love.init()
 			usedpiscale = c.window.usedpiscale,
 			x = c.window.x,
 			y = c.window.y,
+			title = c.window.title or "Untitled"
 		}), "Could not set window mode")
-		love.window.setTitle(c.window.title or c.title)
```
slime73: Could this line be moved above the setMode call, instead of having a bunch of changes to setMode's code?

src/scripts/boot.lua
```diff
@@ -550,8 +550,8 @@ function love.init()
 			usedpiscale = c.window.usedpiscale,
 			x = c.window.x,
 			y = c.window.y,
+			title = c.window.title or "Untitled"
```
slime73: In order for any changes to boot.lua to have any effect, you'll need to run the `auto.lua` script and include the changes to `boot.lua.h` in this review.

```
cd src/scripts
lua auto.lua boot
```

src/scripts/boot.lua
namtabmai: Ah yeah, of course!

Much cleaner solution.

src/scripts/boot.lua
namtabmai: done

PR #1578, opened 2020-05-20
Make getTime() relative to first invocation

pfirsich:
As we have talked about a bit on the Discord server, I made getTimer relative to it's first invocation (which happens at the start of love.run), so that there are less precision issues (and also more predictable behaviour).
I did some investigation of whether, if starting from 0, floats would be enough for most tasks and according to the godbolt I shared on the server too (https://godbolt.org/z/M4HadW) it seems that even for a week of runtime, the precision is still pretty good.
I did take a minute or two to understand all the macro stuff to handle the different platforms and moved some stuff around to make it easier to understand, but I realize this could possibly be a change too big to justify the little feature impact, so if you don't like it, I can change that back.
I tested by printing love.time.getTime() in love.load to the console and to the screen in love.draw and it seems the first value returned is about 0.2 on my system, so maybe the first call is not in love.run. I don't think that's a problem per se, but it might make my comments inaccurate.
I also ran a random program I was also working on and that worked fine. Maybe more testing should be done?

### DIFF
--- a/src/modules/timer/Timer.cpp
+++ b/src/modules/timer/Timer.cpp
@@ -24,6 +24,7 @@
 #include "common/delay.h"
 #include "Timer.h"
 
+#include <iostream>
 #if defined(LOVE_WINDOWS)
 #include <windows.h>
 #elif defined(LOVE_MACOSX) || defined(LOVE_IOS)
@@ -134,11 +135,10 @@ double Timer::getTime()
 {
 	static const timespec start = getTimeAbsolute();
 	const timespec now = getTimeAbsolute();
-	const timespec rel = timespec {
-		now.tv_sec - start.tv_sec,
-		now.tv_nsec - start.tv_nsec
-	};
-	return (double) rel.tv_sec + (double) rel.tv_nsec / 1.0e9;
+	// tv_sec and tv_nsec should be signed on POSIX, so we are fine in just subtracting here.
+	const long sec = now.tv_sec - start.tv_sec;
+	const long nsec = now.tv_nsec - start.tv_nsec;
+	return (double) sec + (double) nsec / 1.0e9;
 }
 
 #elif defined(LOVE_MACOSX) || defined(LOVE_IOS)
--- a/src/modules/timer/Timer.cpp
+++ b/src/modules/timer/Timer.cpp
@@ -182,7 +182,7 @@ double Timer::getTime()
 	const LARGE_INTEGER now = getTimeAbsolute();
 	LARGE_INTEGER rel;
 	rel.QuadPart = now.QuadPart - start.QuadPart;
-	return rel.QuadPart / freq.QuadPart;
+	return (double) rel.QuadPart / (double) freq.QuadPart;
 }
 
 #endif
--- a/src/modules/timer/Timer.cpp
+++ b/src/modules/timer/Timer.cpp
@@ -35,15 +35,6 @@
 #include <sys/time.h>
 #endif
 
-#if defined(LOVE_LINUX)
-static inline double getTimeOfDay()
-{
-	timeval t;
-	gettimeofday(&t, NULL);
-	return (double) t.tv_sec + (double) t.tv_usec / 1000000.0;
-}
-#endif
-
 namespace love
 {
 namespace timer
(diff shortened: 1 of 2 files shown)


pfirsich:
My sad attempt to make it more readable was incredibly misguided and not very smart. I learned something and changed it back to how it was.

src/modules/timer/Timer.cpp
```diff
@@ -156,5 +156,11 @@ double Timer::getTime()
 #endif
 }
 
+double Timer::getTime()
```
slime73, line 159: I feel like it might make sense to call `getTime` in the Timer module constructor, so it has a consistent start point (for people who load the timer module at least). Thoughts?

src/modules/timer/Timer.cpp
```diff
@@ -156,5 +156,11 @@ double Timer::getTime()
...
+{
+	static const double start = getTimeAbsolute();
+	return getTimeAbsolute() - start;
```
slime73, line 162: In the future I'll probably want to restructure this so `getTimeAbsolute` will return the platform-specific integer values directly, which would maximize precision of the final floating point getTime result - but that'd be a much more complicated change and this is a good change to have regardless, I think.

src/modules/timer/Timer.cpp
pfirsich: I thought about doing that, but I wasn't exactly sure where (the Timer module constructor makes a lot of sense). I decided that it probably doesn't matter technically. In regards to the precision problem the difference is small (and probably even a tiny bit better to have it start at first invocation). But maybe one of them is easier to explain and track. I trust your judgment in this.

pfirsich: I just notice that the Timer module creates an instance of the Timer class, which calls getTime in its constructor. So do I just have do adjust the comments?

slime73: Oh, I forgot about that... in that case, just changing the comment to not reference love.run will be sufficient, yeah

src/modules/timer/Timer.cpp
pfirsich: I don't know if there is ever a need to use getAbsoluteTime (this is why I made it a static class function, because maybe there is). If you did it that way, you would not give that possibility anymore. But I don't think need the absolute version really.
After I un-refactored that conditional compilation stuff I thought it would be nice to have a single source file for each platform and just define getTime in there. And considering with different return types you need to wrap the whole functions (and definitions in the header) with #ifdef maybe it makes sense to just have getTime in the header and a separate source file for each platform? Then you can inline getTimerPeriod too and do `counter / frequency` directly instead of `counter * (1/frequency)`. I could also do that in the same file.

pfirsich:
I tried out my test program on Windows as well and that works.

src/modules/timer/Timer.cpp
```diff
@@ -109,52 +100,92 @@ double Timer::getAverageDelta() const
...
+	return getTimeOfDay();
+}
+
+double Timer::getTime()
+{
+	static const timespec start = getTimeAbsolute();
+	const timespec now = getTimeAbsolute();
+	const timespec rel = timespec {
+		now.tv_sec - start.tv_sec,
+		now.tv_nsec - start.tv_nsec
```
slime73, line 139: I think this math won't work in all situations, eg a start time of 5.8 and a "now" time of 6.1.

I found this page after a quick google which has a fairly complicated bit of code at the bottom for subtracting *spec values: http://elias.rhi.hi.is/libc/Elapsed-Time.html . Maybe something cleaner exists elsewhere though.

pfirsich: When I wrote this I just checked the POSIX docs for `timespec` and they said that `timespec.tv_nsec` is a `long` (i.e. signed). Same with the C11 standard, so I thought it was fine. And on POSIX `tv_sec` (`time_t`) should be signed too. It is true however that rel is not a valid timespec object, because `tv_nsec` might be negative. I should probably add a comment that we don't care about rel being a somewhat malformed timespec or rather use two variables with specifically signed types instead.

I think that code you linked is concerned with turning it into a valid timespec. Or am I misunderstanding and it also doesn't work if those fields are both signed?

slime73: Oh, I didn't realize they were signed. Looks good in that case (unless I'm missing something myself with fractional math).

PR #1607, opened 2020-08-13
Update boot.lua

Cartread:
Removing the alpha variable allows anyone who overwrote the setColor function to see errorhandler's print.
e.g. I'm overwriting the setColor function to accommodate 0-255 values and the error message is being printed with 1/255 alpha (invisible).
Removing the 1 will print the message in black.

P.S. I'm currently overwriting errorhandler for this, but I assume others are also overwriting setColor and see no error messages.


MikuAuahDark:
Changes in boot.lua doesn't take effect immediately. Run `auto.lua` script to apply the changes as follows.
```
lua auto.lua boot.lua
```

slime73:
I'm fine with this change (after auto.lua is run) because it doesn't add any code, but in general I don't think it's a good idea for love's codebase to go out of its way to work with unsupported color ranges.

Cartread:
I agree with you Alex.  It's not a fault of LOVE, but I wanted to share the predicament I unfolded yesterday.  I didn't know what happened to the error text, but it was being printed invisibly for 2 years! LOL.  I just used console, and then finally investigated yesterday.

> Changes in boot.lua doesn't take effect immediately. Run `auto.lua` script to apply the changes as follows.
> 
> ```
> lua auto.lua boot.lua
> ```
> 
> Then rebuild.

I'm afraid I'm in over my head.  I'm new to GitHub and cloud based software.  I don't know where to add that line.
Please help or if anyone wants to do a different pull request, I'd be happy with that.

slime73:
`auto.lua` is a script you run locally, you don't need to modify it.

Running it with a Lua interpreter (and passing boot.lua in as an argument) will update the `boot.lua.h` file to match `boot.lua`'s contents.

PR #1609, opened 2020-08-14
make love lua 5.4 conform

nikeinikei:


src/libraries/lua53/lutf8lib.c
```diff
@@ -309,8 +309,12 @@ int luaopen_luautf8 (lua_State *L) {
       lua_setfield(L, -2, l->name);
     }
   }
+#if LUA_VERSION_NUM >= 504
+  lua_pushlstring(L, UTF8PATT, sizeof(UTF8PATT) / sizeof(char) - 1);
```
slime73, line 313: is lua_pushliteral gone or broken somehow? I still see it in the 5.4 documentation

slime73: Now that I look closer, I can imagine the embedded null byte in the pattern might cause problems - that null byte is in 5.2 as well, should this codepath be run all the time instead of behind a version #if ?

nikeinikei: No you're right, it still exists in lua 5.4, but love would crash on this line 
**https://github.com/love2d/love/blob/master/src/scripts/boot.lua#L683**
with the lua_pushliteral variant, when a lua error occured in the game code (utf8.charpattern was malformed).
The lua_pushlstring variant comes from the lua 5.4 lutf8lib.c.

nikeinikei: I checked the 5.3 version of the luaopen_luautf8 function, and it also uses the lua_pushlstring variant. Now I'm wondering where the lua_pushliteral code came from? Was there a reason it was changed?

slime73: I don't think there's a good reason why it originally used pushliteral, aside from saving a tiny bit of code when the file only worked with Lua 5.1. Making it use the pushlstring version all the time would be good IMO.

PR #1613, opened 2020-08-16
Box2d update to 2.4.1

thegrb93:
Will probably work some more tomorrow on it


thegrb93:
I heard there was some modifications to box2d originally for love2d, so those will need to be added in this as well.

slime73:
I think this should be based on the 12.0-development branch rather than master, since master right now is mostly for fixes and small additions to 11.x.

slime73:
> I heard there was some modifications to box2d originally for love2d, so those will need to be added in this as well.

From what I remember, we just have one modification - changing b2Assert in `Common/b2Settings.h` to throw a love C++ exception (via a new `loveAssert` function defined in b2Settings.cpp), rather than calling C's assert().

thegrb93:
Ok, that shouldn't be difficult. Another task is to implement one-sided edge vs two-sided edge, and also ChainShape and EdgeShape's ghost vertices are no longer optional.

thegrb93:
I should also rename the wrapper functions to their new box2d names. Also looks like Linux is failing to compile due to folder name casing.

thegrb93:
Does anyone know whats going on here? It looks like a bug
https://github.com/love2d/love/blob/master/src/modules/physics/box2d/Physics.cpp#L210-L213

thegrb93:
Nvm, it just decrements the reference counter

thegrb93:
Looks like the CI is having a problem with the infinite function. For some reason its using cmath instead of math.h.

thegrb93:
It compiles now but its having some kind of problem linking to liblove.so. I'll go ahead and ready for review.

thegrb93:
Can't figure out how to get the linux build working. Someone said the autoconf is what does it, but I wasn't about to find any config scripts using the wrong box2d paths.

thegrb93:
Do you know? @slime73

slime73:
> Can't figure out how to get the linux build working. Someone said the autoconf is what does it, but I wasn't about to find any config scripts using the wrong box2d paths.

Sorry for taking so long to reply. I'm not too familiar with the Linux build setup but my guess is it has to do with the case change from Box2D to box2d - I see the `genmodules` script iterates over folders in the src/libraries and uses the folder names, and `configure.ac` references `enable_library_Box2D`. Maybe it should be changed to `enable_library_box2d`?

thegrb93:
Thanks. I'll try that. Also I'm going to wait a bit because there's some more good updates on the way to box2d.

thegrb93:
@slime73 Looks like that didn't work. I'd have to install linux and debug through it myself rather than using the CI to figure out what's wrong.

slime73:
The build failure is different this time, at least - before that commit it was failing during linking because box2d symbols couldn't be found (presumably because the box2d library wasn't being built), now it's failing while compiling b2_island.cpp with an error about `dynamics/b2_contact_solver.h` not being found via #include.

slime73:
I think this is another casing issue - in the erincatto/box2d repository, the subfolders in `src` are all lower case (`collision'`, `common`, `dynamics`, `rope`), whereas they're all capitalized (`Collision`, `Common`, `Dynamics`, `Rope`) in both the older version of box2d that's currently in love's repo, and the new one in your PR.

thegrb93:
Oh good catch. Bad assumption on my part.

thegrb93:
Yey Linux is working now. I imagine osx will need to be updated too (that I most definitely can't do). I think this should wait until https://github.com/erincatto/box2d/issues/645 is fixed and then I'll update box2d again.

slime73:
> Yey Linux is working now. I imagine osx will need to be updated too (that I most definitely can't do). I think this should wait until erincatto/box2d#645 is fixed and then I'll update box2d again.

Sounds good. Don't worry about macOS, I'll update the Xcode project file once this goes in.

thegrb93:
Looks like I can update again now. Will start that.

thegrb93:
Will update to https://github.com/erincatto/box2d/releases/tag/v2.4.1

thegrb93:
Need to add missing
```cpp
void loveAssert(bool test, const char *teststr)	
{	
	if (!test)	
		throw love::Exception("Box2D assertion failed: %s", teststr);	
}
```

slime73:
Thanks for doing all the work to get the update working!

thegrb93:
No problem, hopefully everything still works smoothly.

PR #1622, opened 2020-09-17
Add Body.getKinematicState and Body.setKinematicState

thegrb93:
Allows getting/setting a body's full kinematic state in a single call. My use case is

```lua
function myentity:think()
    self.x, self.y, self.a, self.dx, self.dy, self.da = self.body:getState()
    ...
end
```

### DIFF
--- a/src/modules/physics/box2d/Body.cpp
+++ b/src/modules/physics/box2d/Body.cpp
@@ -112,7 +112,7 @@ float Body::getAngularVelocity() const
 	return body->GetAngularVelocity();
 }
 
-void Body::getState(b2Vec2 &pos_o, float &a_o, b2Vec2 &vel_o, float &da_o) const
+void Body::getKinematicState(b2Vec2 &pos_o, float &a_o, b2Vec2 &vel_o, float &da_o) const
 {
 	pos_o = Physics::scaleUp(body->GetPosition());
 	a_o = body->GetAngle();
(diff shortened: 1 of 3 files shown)


slime73:
This only gets/sets a subset of the Body's state so I think the name is a bit misleading.

> My use case is

I'm not 100% sure I understand that - is it just to remove a couple method calls? What's the larger purpose in removing those calls?

thegrb93:
Maybe I should rename to get/setKinematicState. 

It would optimize
```lua
function Collider:getState()
    local x, y = self:getPosition()
    local a = self:getAngle()
    local dx, dy = self:getLinearVelocity()
    local da = self:getAngularVelocity()
    return x, y, a, dx, dy, da
end
```

slime73:
What sort of optimization numbers does it have, both in terms of relative performance and in a real-world situation?

Usually when a love method call itself is expensive (but the work it does is cheap) and it's used in inner loops, the first thing I think of is if the FFI can be used. With physics it gets a bit ambiguous as to whether adding FFI versions of a few methods will help or hurt though I guess, because performance gets wrecked when FFI code can't be JIT compiled - which happens when it's in the same codepath as a traditional C binding function or other NYI bit of code.

thegrb93:
I measured it to be about 63.5% faster. For 1000 bodies on my 4.6Ghz cpu, that's 115us before and 42us after just for fetching the body's states.

thegrb93:
I forgot to ask, is there a changelog that needs to be updated or does the wiki need to be updated?

slime73:
We don't have a changelog in the repository or wiki yet for 12.0 - we'll make both once it gets a bit closer to release.

PR #1623, opened 2020-09-17
Pass preprocessor defines in newShader

nagolove:


src/modules/graphics/wrap_Graphics.cpp
```diff
@@ -1313,6 +1313,10 @@ static int w_getShaderSource(lua_State *L, int startidx, bool gles, std::string
 	bool has_arg1 = lua_isstring(L, startidx + 0) != 0;
 	bool has_arg2 = lua_isstring(L, startidx + 1) != 0;
 
+    bool has_defines = false;
+    if (has_arg1 && !has_arg2) has_defines = lua_istable(L, startidx + 1) != 0;
+    if (has_arg1 && has_arg2) has_defines = lua_istable(L, startidx + 2) != 0;
+
```
MikuAuahDark, line 1319: Please indent with 4 tabs as described here https://love2d.org/wiki/Code_Style

nagolove: Done.

src/modules/graphics/wrap_GraphicsShader.lua
```diff
@@ -324,9 +324,16 @@ local function createShaderStageCode(stage, code, lang, gles, glsl1on3, gammacor
 		GLSL.FUNCTIONS,
 		GLSL[stage].FUNCTIONS,
 		custom and GLSL[stage].MAIN_CUSTOM or GLSL[stage].MAIN,
-		((lang == "glsl1" or glsl1on3) and not gles) and "#line 0" or "#line 1",
-		code,
 	}
+	if defines then
+		for k, v in pairs(defines) do
+			if type(k) == "string" and (type(v) == "string" or type(v) == "number" or type(v) == "boolean") then
+				table.insert(lines, string.format("#define %s %s", k, tostring(v)))
```
slime73, line 331: I think true/false should be converted to 1 and 0, since you can't do `#if true` in GLSL to my knowledge (whereas `#if 1` and `#if 0` work).

src/modules/graphics/wrap_GraphicsShader.lua
```diff
@@ -324,9 +324,16 @@ local function createShaderStageCode(stage, code, lang, gles, glsl1on3, gammacor
 		GLSL.FUNCTIONS,
 		GLSL[stage].FUNCTIONS,
 		custom and GLSL[stage].MAIN_CUSTOM or GLSL[stage].MAIN,
-		((lang == "glsl1" or glsl1on3) and not gles) and "#line 0" or "#line 1",
-		code,
 	}
+	if defines then
+		for k, v in pairs(defines) do
```
slime73, line 329: Iterating over the whole table that's passed in and treating every single field as a #define prevents the table from having other uses. I'm not sure if the table *should* have other uses, or if there's anything else it could be used for, but it's worth thinking about before locking ourselves out of that possibility.

slime73:
I'd like this to be based on the 12.0-development branch instead, since the shader parsing code has been rewritten there to use C++ instead of Lua.

slime73 (changes requested):
(Just letting github formally know I already informally requested changes)

slime73:
This is now in the 12.0-development branch as of commit 0fa00e0 - with slightly different syntax.

PR #1637, opened 2020-10-11
System.cpp: Do not require posix_spawn on "Linux"

glebm:
Many libc implementations do not implement posix_spawn.
Allow love2d to build without it.

### DIFF
--- a/src/modules/system/System.cpp
+++ b/src/modules/system/System.cpp
@@ -39,7 +39,18 @@
 #if defined(LOVE_ANDROID)
 #include "common/android.h"
 #elif defined(LOVE_LINUX)
+
+#ifdef __has_include
+#if __has_include(<spawn.h>)
+#define LOVE_HAS_POSIX_SPAWN
+#endif
+#endif
+
+#ifdef LOVE_HAS_POSIX_SPAWN
 #include <spawn.h>
+#else
+#include <unistd.h>
+#endif
 #endif
 
 namespace love
@@ -70,10 +81,12 @@ std::string System::getOS() const
 #endif
 }
 
+#ifdef LOVE_HAS_POSIX_SPAWN
 extern "C"
 {
 	extern char **environ; // The environment, always available
 }
+#endif
 
 bool System::openURL(const std::string &url) const
 {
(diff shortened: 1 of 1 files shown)


src/modules/system/System.cpp
```diff
@@ -104,10 +117,18 @@ bool System::openURL(const std::string &url) const
...
 	const char *argv[] = {"xdg-open", url.c_str(), nullptr};
 
+#ifdef LOVE_HAS_POSIX_SPAWN
 	// Note: at the moment this process inherits our file descriptors.
 	// Note: the below const_cast is really ugly as well.
 	if (posix_spawnp(&pid, "xdg-open", nullptr, nullptr, const_cast<char **>(argv), environ) != 0)
 		return false;
+#else
+	pid = fork();
+	if (pid == 0) {
```
slime73, line 127: The `{` should be in its own line, as per https://love2d.org/wiki/Code_Style

src/modules/system/System.cpp
```diff
@@ -39,7 +39,18 @@
 #if defined(LOVE_ANDROID)
 #include "common/android.h"
 #elif defined(LOVE_LINUX)
+
+#ifdef __has_include
```
slime73, line 43: What's the availability of `__has_include` like? Is it possible for `posix_spawnp` to exist but `__has_include` isn't supported?

slime73:
> Many libc implementations do not implement posix_spawn.

Which Linux systems in particular that love would otherwise would build on are prevented from being supported by the current `posix_spawn` code?

src/modules/system/System.cpp
glebm: Yes, they're orthogonal: `__has_include` is a C++17 language feature, `posix_spawn` is a function available in some libc implementations.

glebm:
> Which Linux systems in particular that love would otherwise would build on are prevented from being supported by the current `posix_spawn` code?

All of the handheld devices that run OpenDingux-like systems. Example device: RetroGame RG350P

They use versions of uClibc(-ng) as their libc and it doesn't have `posix_spawnp`.

src/modules/system/System.cpp
glebm: clang and gcc have had `__has_include` for a long time (way before it was standardized in C++17)

glebm:
Another big target is OpenWRT router firmware (also uClibc). Doesn't everyone want to run love2d games on their router? :smiley:

MikuAuahDark:
While this change is _probably_ okay, running LOVE at platforms you mentioned may be hard, if not impossible because those platforms most likely lack the power and memory to run LOVE.

glebm:
@MikuAuahDark That's not true, these devices have been using an older version of Love since 2014!

RG350 has 512MiB of RAM and a GPU

src/modules/system/System.cpp
glebm: Done

PR #1645, opened 2020-12-05
LÖVE on Windows 10 ARM patches (step 2)

MikuAuahDark:
There's not much to patch in here actually. All the commit message are self-explanatory regarding the changes.

However, when this eventually be merged, I need to think how we should expose the GitHub Actions artifacts for ARM64 build since it requires ANGLE to function.

Any reviews and pointers regarding above point are welcome.

Before merging this, consider merging love2d/megasource#2 first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,8 +65,6 @@ if(POLICY CMP0072)
 	cmake_policy(SET CMP0072 NEW)
 endif()
 
-find_package(OpenGL)
-
 if(MEGA)
 	# LOVE_MSVC_DLLS contains runtime DLLs that should be bundled with the love
 	# binary (in e.g. the installer). Example: msvcp140.dll.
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -88,7 +86,6 @@ if(MEGA)
 	set(LOVE_LINK_DIRS ${SDL_LINK_DIR})
 
 	set(LOVE_LINK_LIBRARIES
-		${OPENGL_gl_LIBRARY}
```
slime73, line 91: This change doesn't break anything?

MikuAuahDark: Quick test with my laptop, LOVE launches fine without it. Not sure about Linux though, but I think autotools is preferred there.

MikuAuahDark:
On related note, since LOVE in Windows 10 ARM doesn't use LuaJIT, the `bit` library is probably needs to be sourced from https://bitop.luajit.org/ where it loads the bundled `bit` library if (and only if) Lua 5.1 is used.

slime73:
I'm also not sure about the appveyor 64 bit build (failing on 12.0-development too), but at least fixing it probably won't involve changing this PR so I think it's fine to merge.

MikuAuahDark:
I had seen the appveyor build failing prior merging the megasource commits so I think it's something to do with appveyor.

awfulcooking:
LuaJIT now supports Windows on ARM :-) https://github.com/LuaJIT/LuaJIT/issues/593

MikuAuahDark:
love2d/megasource#12 is needed to fully utilize it though.

PR #1646, opened 2020-12-09
adds iOS audio interruption handling and recovery

cigumo:
iOS audio interruptions (eg. alarm, or phone call ringer) disable the OpenAL context. 
This code handles the interruptions and restores the OpenAL context when the interrupt ends or the application is brought to the foreground. 

NOTE: I'm not happy with adding two methods to the Audio class to pause/restore the OpenAL context. Maybe you can think of a more elegant way to do it.

### DIFF
--- a/src/common/ios.mm
+++ b/src/common/ios.mm
@@ -28,6 +28,8 @@
 #import <AudioToolbox/AudioServices.h>
 #import <AVFoundation/AVFoundation.h>
 
+#include "modules/audio/Audio.h"
+
 #include <vector>
 
 #include <SDL_events.h>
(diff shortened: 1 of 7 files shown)


src/common/ios.mm
```diff
@@ -200,6 +202,53 @@ static int dropFileEventFilter(void *userdata, SDL_Event *event)
...
+	__strong static id _shared = nil;
+	dispatch_once(&pred, ^{
+			_shared = [[self alloc] init];
+		});
+	return _shared;
+}
+
+- (void)audioSessionInterruption:(NSNotification *)note
+{
+	@synchronized (self) {
```
slime73, line 223: Code style should have braces on new lines (and no braces when there's a single line in the scope).

src/modules/audio/openal/Audio.cpp
```diff
@@ -293,6 +305,17 @@ std::vector<love::audio::Source*> Audio::pause()
 	return Source::pause(pool);
 }
 
+void Audio::pauseContext()
+{
+	alcMakeContextCurrent(nullptr);
```
slime73, line 310: Does anything need to be done with the audio pool thread? I'm not sure whether it will do the right thing after this happens (eg will any audio be lost from streaming sources?)

cigumo: My tests so far on iOS 12 and 14 don't show any issues. Audio resumes properly when the context is restored. 
I'm going to test some more with streaming loopable sources for interruptions longer than the sound duration.

cigumo: I've tested it with streaming loopable sources and it works fine.

PR #1658, opened 2020-12-31
Support for >2GB DroppedFile in desktops.

MikuAuahDark:
I haven't tested Linux yet, probably tomorrow. That's why I marked this as draft PR.

### DIFF
--- a/src/modules/filesystem/DroppedFile.cpp
+++ b/src/modules/filesystem/DroppedFile.cpp
@@ -106,14 +106,25 @@ bool DroppedFile::isOpen() const
 
 int64 DroppedFile::getSize()
 {
+	int fd = file ? fileno(file) : -1;
+
 #ifdef LOVE_WINDOWS
+	
+	struct _stat64 buf;
 
-	// make sure non-ASCII filenames work.
-	std::wstring wfilename = to_widestr(filename);
+	if (fd != -1)
+	{
+		if (_fstat64(fd, &buf) != 0)
+			return -1;
+	}
+	else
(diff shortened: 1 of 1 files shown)


src/modules/filesystem/DroppedFile.cpp
```diff
@@ -173,12 +190,22 @@ int64 DroppedFile::tell()
...
 bool DroppedFile::seek(uint64 pos)
 {
-	return file != nullptr && fseek(file, (long) pos, SEEK_SET) == 0;
+	return file != nullptr &&
+
+#ifdef LOVE_WINDOWS
+	_fseeki64(file, (int64) pos, SEEK_SET) == 0;
+#else
+	fseeko(file, (off_t) pos, SEEK_SET) == 0;
+#endif
```
qaisjp, line 208:
```suggestion
	if (file == nullptr)
		return false;

#ifdef LOVE_WINDOWS
	return _fseeki64(file, (int64) pos, SEEK_SET) == 0;
#else
	return fseeko(file, (off_t) pos, SEEK_SET) == 0;
#endif
```

Splitting this `return` on multiple lines feels kinda unexpected (to me)

MikuAuahDark: I think you're right.

MikuAuahDark:
Test in Windows is good. Linux probably needs `_FILE_OFFSET_BITS=64` to be set (dunno if this is set at the autotools). Android has no link errors (but it's not supported anyway).

PR #1659, opened 2021-01-02
Expand love.filesystem APIs

### LINKED ISSUE #1219: Easy access from love2d to ios file sharing directory
**[Original report](https://bitbucket.org/rude/love/issue/1214) by Sylvain Boussekey (Bitbucket: [VertPingouin](https://bitbucket.org/VertPingouin), GitHub: [VertPingouin](https://github.com/VertPingouin)).**

It would be useful to be able to access some file put in the file sharing directory via itunes. It would immensly expands possibility in ios.

Maybe the sandboxed nature of ios app prevent accessing it directly, but being able, at least, to copy some files at runtime to be used in game (like custom levels or save files) would be a  great addition.

### LINKED ISSUE #1360: Mount any folder
**[Original report](https://bitbucket.org/rude/love/issue/1355) by Pablo Mayobre (Bitbucket: [PabloMayobre](https://bitbucket.org/PabloMayobre), GitHub: [PabloMayobre](https://github.com/PabloMayobre)).**

Currently LÖVE allows you to mount dropped folders, zip files, and the SourceBaseDirectory when fused.

Mounting arbitrary directories is not allowed by LÖVE.

A use case for this feature would be, remembering the folder that the user dropped in an old session of the game, by saving the filepath of the folder to a file and in later sessions mounting that same location.

Note that this would be read-only and that `love.filesystem.mount` returns `true` in case of `success`, so it could return false if the directory doesn't exist or is not accessible.

This would allow to list, read and get info on the directory contents. Writing could be achieved through Lua `io` functions with the already known downsides that come with it.

slime73:
This change adds several new love.filesystem APIs for reading and writing in directories outside of the game source and save paths.

As part of this change, PhysFS had to be extended with two new APIs (PHYSFS_mountRW and PHYSFS_canUnmount).

* Add `love.filesystem.mountFullPath` and `love.filesystem.unmountFullPath`:
```Lua
success = love.filesystem.mountFullPath(fullpath, mountpoint, permissions, appendtopath)
success = love.filesystem.unmountFullPath(fullpath)
```
* `permissions` is an enum that can be `"read"` or `"readwrite"`. Operating systems may make more full (platform-dependent) paths readable than writable.

* Add `love.filesystem.mountCommonPath`, `love.filesystem.unmountCommonPath`, `love.filesystem.getFullCommonPath`.
```Lua
success = love.filesystem.mountCommonPath(commonpath, mountpoint, permissions, appendtopath)
success = love.filesystem.unmountCommonPath(commonpath)
fullpath = love.filesystem.getFullCommonPath(commonpath)
```

* `CommonPath` is a new enum containing several values:
```Lua
"userhome" -- The user's home directory.
"userappdata" -- The user's appdata directory.
"userdesktop" -- The user's desktop directory.
"userdocuments" -- The user's documents directory
"appdocuments" -- Similar to the save directory, a subfolder in the user's documents directory based on the game's identity string.
```

* Add `readonly` boolean field to the table returned by `love.filesystem.getInfo`.

Resolves #1360 
Resolves #1219


MikuAuahDark:
While this is good change, I have several notes:

1. `love.filesystem.mountFullPath` won't work in Android 10 and later due to scoped storage restrictions (unless using legacy external storage). This should work in Android 9 and earlier as long as the storage permission is granted by user. There's, a way, [to make it work in Android 11 and later][MANAGE_EXTERNAL_STORAGE] but this have high risk of LOVE app being rejected from Play Store.
2. Android doesn't have "desktop" directory, so `love.filesystem.mountCommonPath` with `"userdesktop"` `CommonPath` won't work there.

And several questions:

1. How LOVE can handle in case certain `CommonPath` is not supported at all in certain platforms?
2. For Android, `"userdocuments"` `CommonPath` enum won't work in (and only in) Android 10. Android 9 and earlier has no scoped storage restrictions, while Android 11 adds support of opening documents directory with NDK. How it should handle these cases?

I have suggestion, what about additional `"temp"` `CommonPath` enum which mounts the temp/cache directory?

[MANAGE_EXTERNAL_STORAGE]: https://developer.android.com/reference/android/Manifest.permission#MANAGE_EXTERNAL_STORAGE

slime73:
> love.filesystem.mountFullPath won't work in Android 10 and later due to scoped storage restrictions

Yeah, that API isn't intended to be used on mobile really (or other sandboxed platforms). Part of the API design for mountFullPath is that it's not guaranteed to succeed because of reasons like that, and code that calls it should be ready for when it doesn't.

> How LOVE can handle in case certain CommonPath is not supported at all in certain platforms?

Same as above, mountCommonPath is meant to return false in situations where the platform doesn't allow it or the path doesn't conceptually fit with the platform.

> I have suggestion, what about additional "temp" CommonPath enum which mounts the temp/cache directory?

I had a partial implementation of that in one commit, but I decided to temporarily (heh) remove it to focus on getting everything else working first. I'll probably do a full implementation of it before 12.0 is released.

src/modules/filesystem/Filesystem.cpp
```diff
@@ -120,30 +194,32 @@ std::string Filesystem::getExecutablePath() const
...
+	{ "userdocuments", Filesystem::COMMONPATH_USER_DOCUMENTS },
 }
+STRINGMAP_CLASS_END(Filesystem, Filesystem::CommonPath, Filesystem::COMMONPATH_MAX_ENUM, commonPath)
 
-std::vector<std::string> Filesystem::getConstants(FileType)
+STRINGMAP_CLASS_BEGIN(Filesystem, Filesystem::MountPermissions, Filesystem::MOUNT_PERMISSIONS_MAX_ENUM, mountPermissions)
 {
-	return fileTypes.getNames();
+	{ "read",      Filesystem::MOUNT_PERMISSIONS_READ      },
+	{ "readwrite", Filesystem::MOUNT_PERMISSIONS_READWRITE },
```
qaisjp, line 220: should this be `r` and `rw` in line with e.g. `io.open`?

slime73: I'm not actually a big fan of using single letter enums - io.open just mirrors what the C fopen API uses, so I don't think it should be as much of a guide compared to the style of other love enums

rgrams:
Why add new functions to the API instead of simply removing the restrictions on love.filesystem.mount?

rgrams:
Got it. Fair enough. Thanks!

PR #1674, opened 2021-02-18
Add MIDI playback support using native APIs (Windows/macOS)

Ruin0x11:
I was wanting to be able to have native MIDI playback in LÖVE for some time for one of my projects (as it's a source port that needs backwards compatibility), so I tried adding it to the codebase about a year ago. It more or less worked, but because LÖVE was hosted on Bitbucket at the time I was constantly fighting Mercurial to try to make the changes contributable under Git, which I'm more familir with. Eventually I ended up shelving it out of exhaustion. Now that LÖVE's repo has moved to Git, I can maybe make some progress on this.

This change contains a new audio decoder for MIDI that uses native MIDI streaming APIs on Windows, CoreAudio on macOS, and a copy of TiMidity sourced from zdoom for Linux. It's kind of hackish in its current state, because when using Windows APIs or CoreAudio, there is no actual data sent to the audio buffer. Instead, MIDI playback is handled by a synthesizer running on the machine like VirtualMIDISynth or the default GS Wavetable synth. Because of this I had to add some hacks to the decoder like new virtual methods that get called around stream start/stop and other things like a dummy buffer size to satisfy the OpenAL code (which would otherwise stop the stream as soon as it started because it sees zero data to be sent).

To be honest the TiMidity support is very spotty at present. The things wrong with it are: there's no reverb, so songs get rendered very strangely, it sometimes clips or studders, the code is taken from a very old copy of TiMidity intended for use with SDL_sound, and I'm not even sure the code is compatible with LÖVE's zlib license (dual LGPL/Artistic License). I would be fine with just dropping support for it in this PR.

Other things that need to be addressed include how to set the default MIDI device on Windows (it's currently hardcoded) and some naming convention-related things.


MikuAuahDark:
MIDI files normally played depending on the app use case. It doesn't contain any audio data, only state of which notes are pressed and which notes aren't.

Unfortunately we can't merge this pull request due to licensing issue. Timidity you're using is licensed under LGPL and statically linking to it means LGPL will supersede zlib license LOVE currently using. This also restricts iOS distribution, so it even if somehow the Timidity is compiled as dynamic library, this feature will most likely absent in iOS version. Furthermore, we have no plans on changing LOVE license at this time (and most likely in the future)

Ruin0x11:
That's fine, it would be nice to have MIDI playback on Linux. I'd just remove anything related to TiMidity if there's still interest in MIDI support.

nagolove:
Does it support only MIDI files playback? I would like to have support of MIDI devices such as keyboards and controllers on Windows, Linux and Android.

MikuAuahDark:
MIDI input device is another new level. For Windows, I think you can use FFI (https://docs.microsoft.com/en-us/windows/win32/multimedia/opening-midi-input-devices). Not sure if that's possible in macOS and not sure if Linux has unified API for it.

In short, we won't likely have built-in MIDI input support in the future, please write your own binding for it.

PR #1715, opened 2021-07-08
Compute Shaders

bjornbytes:
This PR adds support for compute shaders.  It's at least in a state where it can do useful things now so I'm opening it for feedback and as a place to discuss synchronization strategies.  There's probably some cleanup to do and bugs to fix.

[Example](https://gist.github.com/bjornbytes/281b54c001060c1af02b1074aca2c333)

API changes

- `love.graphics.newComputeShader`
- `love.graphics.dispatchThreadgroups`
- `Shader:hasStage`
- `Shader:getLocalThreadgroupSize` -- get the local workgroup size declared in the shader
- new `GraphicsFeature`: `compute` -- whether compute shaders are supported
- new `GraphicsLimit`s: `threadgroupsx`, `threadgroupsy`, `threadgroupsz` -- max values of `dispatchThreadgroups`

The synchronization right now just does a full barrier after dispatching a compute shader, which is unacceptable but I want to think more about synchronization strategies.  There's a spectrum of how "perfect" LÖVE can do its synchronization -- it can do various levels of more sophisticated object tracking to try to get more parallelism out of the barriers, and before implementing something fancy I want to go over the tradeoffs/implementation.

I'm terrible at C++ so please rip it apart.  I'll leave some additional comments about the implementation as a review later.


src/modules/graphics/opengl/Graphics.cpp
```diff
@@ -469,6 +469,12 @@ void Graphics::setActive(bool enable)
 	active = enable;
 }
 
+void Graphics::dispatch(int x, int y, int z)
+{
+	glDispatchCompute(x, y, z);
+	glMemoryBarrier(GL_ALL_BARRIER_BITS); // Strawman
```
slime73, line 475: I'm fine with doing a proper synchronization implementation after this change is merged rather than before

src/modules/graphics/Shader.cpp
```diff
@@ -636,21 +675,26 @@ void Shader::validateDrawState(PrimitiveType primtype, Texture *maintex) const
...
+bool Shader::validateInternal(StrongRef<ShaderStage> stages[], std::string &err, ValidationReflection &reflection)
 {
 	glslang::TProgram program;
 
-	if (vertex != nullptr)
-		program.addShader(vertex->getGLSLangShader());
-
-	if (pixel != nullptr)
-		program.addShader(pixel->getGLSLangShader());
+	for (int i = 0; i < SHADERSTAGE_MAX_ENUM; i++)
```
slime73, line 695: I prefer having braces around multi-line scopes, even if it still works without them

src/modules/graphics/Shader.cpp
```diff
@@ -671,6 +715,10 @@ bool Shader::validateInternal(ShaderStage *vertex, ShaderStage *pixel, std::stri
 		reflection.usesPointSize = vertintermediate->inIoAccessed("gl_PointSize");
 	}
 
+	if (stages[SHADERSTAGE_COMPUTE] != nullptr)
```
slime73, line 718: Same here with braces

src/modules/graphics/Graphics.h
```diff
@@ -436,6 +440,7 @@ class Graphics : public Module
 	ParticleSystem *newParticleSystem(Texture *texture, int size);
 
 	Shader *newShader(const std::vector<std::string> &stagessource);
+	Shader *newComputeShader(std::string source);
```
slime73, line 443: Should be `const std::string &source` so it doesn't make a temporary allocation for the source string every time it's called

src/modules/graphics/Graphics.h
```diff
@@ -145,6 +145,7 @@ class Graphics : public Module
 		FEATURE_INSTANCING,
 		FEATURE_TEXEL_BUFFER,
 		FEATURE_COPY_BUFFER,
+		FEATURE_COMPUTE,
```
slime73, line 148: `FEATURE_GLSL4` covers compute shader availability (after my other comment) so I think a separate enum for compute support isn't needed

src/modules/graphics/opengl/OpenGL.cpp
```diff
@@ -500,6 +503,19 @@ void OpenGL::initMaxValues()
 		maxShaderStorageBufferBindings = 0;
 	}
 
+	if (GLAD_VERSION_4_3 || GLAD_ES_VERSION_3_1 || GLAD_ARB_compute_shader)
```
slime73, line 506: I think we shouldn't bother with `ARB_compute_shader`, since guaranteeing that every system which supports compute shaders in love also supports GL4.3/ES3.1 makes it more reliable to use, and it won't really leave any hardware behind to do so.

src/modules/graphics/wrap_Graphics.cpp
```diff
@@ -3255,6 +3283,16 @@ int w_polygon(lua_State *L)
 	return 0;
 }
 
+int w_dispatchThreadgroups(lua_State* L)
+{
+	Shader *shader = luax_checkshader(L, 1);
+	int x = luaL_optinteger(L, 2, 1);
```
slime73, line 3289: I think it makes sense to require threadgroupsX and keep y/z optional

slime73: Also `optinteger` returns a 64 bit integer (`lua_Integer`) on 64 bit platforms, so there should be an explicit cast down to `int` here.

src/modules/graphics/Shader.cpp
```diff
@@ -671,6 +715,10 @@ bool Shader::validateInternal(ShaderStage *vertex, ShaderStage *pixel, std::stri
...
+		for (int i = 0; i < 3; i++)
+			reflection.localThreadgroupSize[i] = program.getLocalSize(i);
```
slime73, line 720: I don't remember whether glslang / glsl itself will error on its own if `local_size` is omitted or has <= 0 values - I think it's probably worth explicitly outputting an error here if any threadgroup size is <= 0.

src/modules/graphics/Graphics.cpp
```diff
@@ -1065,6 +1081,26 @@ void Graphics::copyBuffer(Buffer *source, Buffer *dest, size_t sourceoffset, siz
...
 
+void Graphics::dispatchThreadgroups(Shader* shader, int x, int y, int z)
+{
+	if (!capabilities.features[FEATURE_COMPUTE])
+		throw love::Exception("Compute shaders are not supported on this system.");
+
+	if (!shader->hasStage(SHADERSTAGE_COMPUTE))
+		throw love::Exception("Only compute shaders can have threads dispatched.");
+
+	if (x > capabilities.limits[LIMIT_THREADGROUPS_X]
```
slime73, line 1092: This should also error if the parameters are <= 0

slime73 (changes requested):
Looks pretty solid overall although I haven't tested it yet, I just have some minor feedback about a few details

PR #1719, opened 2021-08-02
Remove deprecated method love.graphics.isCreated from love.errhand

Sheepolution:
`love.graphics.isCreated` was removed in 0.9.0 according to the wiki.

### DIFF
--- a/src/scripts/boot.lua
+++ b/src/scripts/boot.lua
@@ -646,7 +646,7 @@ function love.errhand(msg)
 		return
 	end
 
-	if not love.graphics.isCreated() or not love.window.isOpen() then
+	if not love.window.isOpen() then
 		local success, status = pcall(love.window.setMode, 800, 600)
 		if not success or not status then
 			return


MikuAuahDark:
I'm not quite sure about this.
https://github.com/love2d/love/blob/1e69a8c4852c2f51849c4ef19bd0d13c4cb3ff05/src/modules/graphics/wrap_Graphics.cpp#L2991

Slime probably can explain?

slime73:
I did a bit of investigating - `love.window.isOpen` was added in version 0.9 when window-related functions were moved to their own module. I think `love.graphics.isCreated` stuck around because it says whether a graphics context is created rather than just whether a window is open - but I *think* love's internal code makes sure a graphics context is always created if a window is open, when the graphics module is active.

So if that's true, the correct thing might be to mark `isCreated` as deprecated (and remove the call from the error handler code as well). In any case the wiki isn't accurate right now.

slime73 (changes requested):
the error handler code moved since this PR was made too, so it will need to be redone (along with marking the function as deprecated), if that's the route people want to take.

That being said, I'm not entirely positive the function has no use in future love versions when other graphics backends are used. I'd have to think more about it (and closely look at my metal code).

Issue #1406, opened 2018-04-12
When other applications are running, videos start to get artifacts and can lock up


slime73:
**[Original report](https://bitbucket.org/rude/love/issue/1401) by Varun Ramesh (Bitbucket: [varunr](https://bitbucket.org/varunr), GitHub: [varunr](https://github.com/varunr)).**

----------------------------------------

When other applications are running (generally video players / recorders), videos in Love start to get artifacts. This is unfortunate, but what's worse is the video will permanently lock up after artifacts accumulate. video:tell() freezes at a certain value and video:isPlaying() always returns true.

When this happens, love.update and love.draw callbacks are still being called, so ui elements still work correctly. However, when the game quits, the process does not exit and must be killed.

I've attached a repro .love. I recommend running it while another video player is playing or OBS is recording.

slime73:
[video_artifacts_lockup.zip](https://github.com/love2d/love/files/4238665/video_artifacts_lockup.zip)

PR #1729, opened 2021-09-21
Break OggDemuxer::readPage loop on EOF.

### LINKED ISSUE #1728: Löve crashes/hangs if trying to make a video from a non video file
To reproduce, just create a random file "a" and try to run the code:
```lua
love.graphics.newVideo("a")
```

MikuAuahDark:
This fixes #1728

### DIFF
--- a/src/modules/video/theora/OggDemuxer.cpp
+++ b/src/modules/video/theora/OggDemuxer.cpp
@@ -43,7 +43,7 @@ OggDemuxer::~OggDemuxer()
 	ogg_sync_clear(&sync);
 }
 
-void OggDemuxer::readPage()
+void OggDemuxer::readPage(bool throweof)
 {
 	char *syncBuffer = nullptr;
 	while (ogg_sync_pageout(&sync, &page) != 1)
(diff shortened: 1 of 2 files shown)


slime73:
The theora decoding code is really hard to follow.. I'll need to test this on a few videos I think (especially ones that have caused problems in the past) before knowing if it's safe to merge

rameshvarun:
This is related to a long standing issue that I initially reported in #1406 and still occurs on master. My understanding:

- In some edge cases we fail to establish sync before the EOF.
- readPage hangs in an infinite loop with file->read always returning 0

This PR doesn't fully fix that problem but I think it's in the right direction. Right now it solves the case where we can't establish sync during initialization.

src/modules/video/theora/OggDemuxer.cpp
```diff
@@ -118,22 +121,23 @@ OggDemuxer::StreamType OggDemuxer::findStream()
...
 		ogg_stream_clear(&stream);
 		ogg_sync_reset(&sync);
 	}
 
-	streamInited = true;
 	while (true)
 	{
 		// If this page isn't at the start of a stream, we've seen all streams
-		readPage();
+		readPage(true);
```
slime73, line 133: Maybe it'd be better for readPage to return something instead of optionally throwing an exception? And then this code could `return STREAM_UNKNOWN` if readPage returns a bad value, or something.

I only suggest it because `TheoraVideoStream`'s constructor which calls `findStream` checks its return value and throws a slightly more detailed exception, already.

PR #1739, opened 2021-10-31
Fix #1707: Font:getWrap() removing final newline charancter

### LINKED ISSUE #1707: Font:getWrap() removes final newline character; doesn't give empty string in last element of table returned.
Here is a love.draw() which illustrates the point.

	function love.draw()
		-- get some text with multiple lines to wrap.
		local text = "line 1, some text to wrap\n"..
		"line 2, some text to wrap\n"..
		"line 3, the next line is intentionally blank"

		-- print the text
		love.graphics.print(text, 50, 50)

		-- now we try with font:getWrap
		local font = love.graphics.getFont()

		-- wrap amount doesn't matter in this case
		width, wrapped = font:getWrap(text, 800)

		-- print each line
		for k,v in ipairs(wrapped) do
			love.graphics.print(k.." "..v, 50, 120 + ((k - 1) * font:getHeight()))
		end

		-- this looks the same:
		-- now we try with an extra \n
		text = text.."\n"
		width, wrapped = font:getWrap(text, 800)
		for k,v in ipairs(wrapped) do
			love.graphics.print(k.." "..v, 50, 190 + ((k - 1) * font:getHeight()))
		end
		-- this prints 3 lines.

		-- One more:
		text = text.."\n"
		width, wrapped = font:getWrap(text, 800)
		for k,v in ipairs(wrapped) do
			love.graphics.print(k.." "..v, 50, 260 + ((k - 1) * font:getHeight()))
		end
		-- this prints 4 lines, and the last one is an empty string.
	end

This makes strings like "xyz" and "xyz\n" return the same values from Font:getWrap(). This only removes the *final* newline character, and you can have other empty new lines at the end of the output, as long as there is something after them (even a newline). This makes me think it isn't just removing whitespace at the end. Adding a space to the end of the string also preserves the newline.

Overall, it's not a huge problem, I can just add another \n onto any string that goes into Font:getWrap() if I need to be sure that there's a newline present at the end.

This is a zip file which has the code above.
[bug.zip](https://github.com/love2d/love/files/6572629/bug.zip)

s-walrus:
Fixes #1707

### DIFF
--- a/src/modules/graphics/Font.cpp
+++ b/src/modules/graphics/Font.cpp
@@ -896,14 +896,11 @@ void Font::getWrap(const ColoredCodepoints &codepoints, float wraplimit, std::ve
 	}
 
 	// Push the last line.
-	if (!wline.cps.empty())
-	{
-		lines.push_back(wline);
+	lines.push_back(wline);
 
-		// Ignore the width of any trailing spaces, for individual lines.
-		if (linewidths)
-			linewidths->push_back(width - widthoftrailingspace);
-	}
+	// Ignore the width of any trailing spaces, for individual lines.
+	if (linewidths)
+		linewidths->push_back(width - widthoftrailingspace);
 }
 
 void Font::getWrap(const std::vector<ColoredString> &text, float wraplimit, std::vector<std::string> &lines, std::vector<int> *linewidths)


src/modules/graphics/Font.cpp
```diff
@@ -896,14 +896,11 @@ void Font::getWrap(const ColoredCodepoints &codepoints, float wraplimit, std::ve
 	}
 
 	// Push the last line.
-	if (!wline.cps.empty())
```
slime73, line 899: We'll need to see if anything special needs to be done for generated newlines (as opposed to explicitly added newlines) there, I suppose. I suspect that was the reason for that conditional in the first place, but I'm not sure if it's actually necessary or not.

Issue #1731, opened 2021-09-27
`RevoluteJoint:getMotorTorque` doesn't work.


FlamingArr:
According to the [docs](https://love2d.org/wiki/RevoluteJoint:getMotorTorque) the `:getMotorTorque()` function takes in no argument and returns a number representing the torque of the joint.
But when calling the function with no args it fails as it expects a number as argument.

Löve's source: https://github.com/love2d/love/blob/master/src/modules/physics/box2d/RevoluteJoint.h#L92

Minimal code example:
```lua
function love.load()
	local world = love.physics.newWorld()
	local body1 = love.physics.newBody(world, 100, 100, "dynamic")
	local body2 = love.physics.newBody(world, 100, 100, "dynamic")
	local revoluteJoint = love.physics.newRevoluteJoint(body1, body2, 100, 100)
	local torque = revoluteJoint:getMotorTorque() -- FAILS
	-- main.lua:6: bad argument #1 to 'getMotorTorque' (number expected, got no value)
end
```

slime73:
Looks like it's just a documentation issue.

megagrump:
[fixed](https://love2d.org/wiki/RevoluteJoint:getMotorTorque)

megagrump:
Would it be acceptable to get `dt` automatically if it wasn't provided in the call? Is changing `invdt` to `dt` a good idea?

slime73:
Yeah I think those are reasonable changes.

slime73:
I think trying to guess the dt might be too error-prone (the frame delta time isn't always used with love.physics, and these APIs might be called before `World:update` rather than always being called during/after, in a given update cycle). So I think this is solved now.

PR #1743, opened 2021-11-21
use dt instead of inv_dt in RevoluteJoint::getMotorTorque

megagrump:
Getting `dt` from `Timer` if not provided in the call is not as straightforward as I thought (not a static method), and I don't want to fuck up my first commit in this codebase by doing something stupid.

Issue #1731

### DIFF
--- a/src/modules/physics/box2d/RevoluteJoint.cpp
+++ b/src/modules/physics/box2d/RevoluteJoint.cpp
@@ -101,8 +101,9 @@ float RevoluteJoint::getMotorSpeed() const
 	return joint->GetMotorSpeed();
 }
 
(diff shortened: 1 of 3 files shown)


thegrb93:
Anybody using getMotorTorque will have their code broken

slime73:
> Getting dt from Timer if not provided in the call is not as straightforward as I thought (not a static method)

I was thinking the last dt passed to World:update could be cached by the world, and RevoluteJoint:getMotorTorque could query that (assuming there's a way to access the owner World from it).

slime73:
> Anybody using getMotorTorque will have their code broken

That's fine, this is for love 12 which has breaking changes.

Issue #1579, opened 2020-05-21
Low precision and range with love.math.noise


megagrump:
`love.math.noise` only gives acceptable results with coordinates < ~200,000. Using larger coordinates shows visible banding/pixellation.

```lua
local noise = love.image.newImageData(800, 600)

for i = 0, 600, 200 do
	noise:mapPixel(function(x, y)
		local n = love.math.noise(i * 1000 + x * .01)
		return n, n, n, 1
	end, i, 0, 200, 150)

	noise:mapPixel(function(x, y)
		local n = love.math.noise(i * 1000 + x * .01, i * 1000 + y * .01)
		return n, n, n, 1
	end, i, 150, 200, 150)

	noise:mapPixel(function(x, y)
		local n = love.math.noise(i * 1000 + x * .01, i * 1000 + y * .01, i * 1000 + y * .01)
		return n, n, n, 1
	end, i, 300, 200, 150)

	noise:mapPixel(function(x, y)
		local n = love.math.noise(i * 1000 + x * .01, i * 1000 + y * .01, i * 1000 + y * .01, i * 1000 + y * .01)
		return n, n, n, 1
	end, i, 450, 200, 150)
end

local img = love.graphics.newImage(noise)

function love.draw()
	love.graphics.draw(img)
end

```

![image](https://user-images.githubusercontent.com/31128870/82528737-bcba7980-9b39-11ea-8c38-e2afa0892446.png)

Also, the range of output values for 1d, 3d and 4d noise is much smaller than it is for 2d noise, resulting in low contrast output.

megagrump:
I wanna fix the aliasing issue at last - the obvious fix is to use `double`s all the way down. I don't think that would be considered a breaking change, would it?

The patent on 3D and 4D noise [expires in a few weeks](https://patents.google.com/patent/US6867776) btw. So we could have SimplexNoise in higher dimensions too in 12.0, if there's no other objection.

zorggn:
Regarding the 3D and 4D ones, can patents be re-filed/renewed? Because if not, then i don't think it'll be an issue come january next year, but if it's a possibility, maybe using OpenSimplex might be a better idea? Not sure if that would have licensing or other issues or not.

slime73:
> the obvious fix is to use doubles all the way down. I don't think that would be considered a breaking change, would it?

I'd consider it breaking because there's a high chance that the output values will change slightly even when the inputs aren't extremely huge. I think it's a good change for 12.0 though.

[...]

PR #1744, opened 2021-11-22
use double precision numbers in love.math.noise

megagrump:
love-noise


slime73:
What's the pic showing? results at different offsets and scales?

megagrump:
Output of repro code in #1579

PR #1746, opened 2021-11-25
added feature 'restart with param override'

stone628:
nogame can restart with dropped directory/file
as first commandline argument


nagolove:
> nogame can restart with dropped directory/file as first commandline argument

I think it useful feature as for user as for developers.

stone628:
There is a bug with this commit that 'love.conf' not invoked using this feature.
I'm still new to love2d src project, and working on this behavior(bug).

slime73:
LÖVE 12 (in the 12.0-development git branch) has a feature similar to this already: there's a new `love.event.restart(optionalarg)` function, and the argument (which can be a string, number, boolean, or a table containing any of those including a table of tables since it's saved as a Variant internally) gets saved and put into a new `love.restart` variable when the game boots back up.

What it can't do right now is cause a different game to be opened, but maybe some Lua code in the boot process (e.g. in [love.boot](https://github.com/love2d/love/blob/12.0-development/src/modules/love/boot.lua#L42)) can be changed to look for a certain value in the restart variable that users aren't likely to unintentionally set, and use that to override the path it looks for the game.

If that's done, then the nogame code (in love 12) could just call `love.event.restart` with that value.

stone628:
Sounds great!

Then I'll keep this code to myself, will share code without bug('love.conf' not called) to forum.
Will do some donation, too. Keep up the good Work!

(I will close this request in a day or two, if it is ok, or moderator can close this for me.)

stone628:
> There is a bug with this commit that 'love.conf' not invoked using this feature. I'm still new to love2d src project, and working on this behavior(bug).

Actually, it was not a bug, but was by design.
I was testing with code taken from nogame.lua, a single file project without conf.lua

PR #1753, opened 2021-12-07
Support for per-character horizontal spacing

stone628:
added Font:getCharSpacing, Font:setCharSpacing
tested with love.graphics.print and Font:getWrap

character spacing is in pixel unit, negative value for letters to be shown closer


slime73:
The general API seems reasonable, although I wonder how it would interact with cursive-style languages like Arabic (which love doesn't fully support yet, but hopefully will in the future). I wonder what other libraries do in that situation?

slime73:
Now that I've been working on support for things like ligatures and other forms of character combining and separating (via Harfbuzz), I think this change isn't compatible with those concepts at the love.graphics Font level.

It makes more sense for it to be done for font files themselves I think - for example ImageFonts already [support](https://love2d.org/wiki/love.graphics.newImageFont) an extra spacing value which gets applied to each character. BMFont definition files have a per-character advance value you can modify, and TrueType fonts can be modified as well.

PR #1757, opened 2021-12-11
Change luaL_loadbuffer chunkname with prefixes.

### LINKED ISSUE #1748: Invalid chunkname values.
When the Lua API expects a `chunkname` argument as with `luaL_loadbuffer` that LÖVE uses, it is actually equivalent to the **source** field documented [here](https://www.lua.org/manual/5.2/manual.html#4.9), not just a chunkname. Without the `=` or `@` prefix, it means plain text source, resulting in the strange `[string "boot.lua"]` output in error messages. Moreover, if LÖVE switchs to LuaJIT 2.1, the profiler will not output plain text source, so it will result in `[string]`, which is meaningless.

E.g. for the embedded `boot.lua`, the correct **chunkname** could be `=[embedded "boot.lua"]` or `@boot.lua`. I don't know about the debug implications of `@`, it may try to look for the file, but `=` is *user-dependent*.

MikuAuahDark:
LOVE's embedded script now shows as `[love "filename.lua"]` and LuaSocket's embedded script shows as `[socket "filename.lua"]`.

Fixes #1748.

### DIFF
--- a/src/common/runtime.cpp
+++ b/src/common/runtime.cpp
@@ -985,7 +985,9 @@ void luax_runwrapper(lua_State *L, const char *filedata, size_t datalen, const c
 	// functions struct pointer as arguments.
 	if (lua_istable(L, -1))
 	{
-		luaL_loadbuffer(L, filedata, datalen, filename);
+		std::string chunkname = std::string("=[love \"") + std::string(filename) + std::string("\"]");
+
+		luaL_loadbuffer(L, filedata, datalen, chunkname.c_str());
 		lua_pushvalue(L, -2);
 		if (ffifuncs != nullptr)
 			luax_pushpointerasstring(L, ffifuncs);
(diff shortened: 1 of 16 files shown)


src/love.cpp
```diff
@@ -187,7 +187,7 @@ static DoneAction runlove(int argc, char **argv, int &retval)
 			lua_rawseti(L, -2, -2);
 		}
 
-		lua_pushstring(L, "embedded boot.lua");
+		lua_pushstring(L, "[love \"boot.lua\"]");
```
slime73, line 190: I did a quick github code search and I see at least some projects/libraries that check for `"embedded boot.lua"` in the args list and do stuff when it's found, so IMO we should change this line for 12.0 rather than 11.4.

slime73:
`luax_runwrapper` should be updated too I guess (to convert its filename argument to the new format), since runwrapper calls `luaL_loadbuffer`.

Issue #1773, opened 2022-02-07
Enforce usage of dots in `require`


MikuAuahDark:
LOVE accidently supports loading modules using `require("path/to/library")` notation. This is not supported in normal Lua environment and it seems some newbies falls into this and sometimes multi-file libraries doesn't handle paths with slashes very well.

Feel free to discuss your opinion about this change. If this is accepted, I think this is good idea to start this enforcement in 12.0.

slime73:
I've been thinking about using love's deprecation mechanism to deprecate slashes for love 12. Maybe if we do that we'll evaluate when to turn the deprecation into a removal based on how many libraries and codebases have been updated.

slime73:
There's also some older discussion about it in #1214 - I suppose keeping it deprecated forever would work around the problem I mentioned there with removal.

slime73:
This is implemented now via deprecation notices, as of 941fe51

PR #1782, opened 2022-03-01
Show deprecation notice when slashes is passed to require.

MikuAuahDark:
As discussed in #1773, it's probably a good idea to show a warning about usage of slashes in `require`, then evalue when to enforce this by throwing Lua error.


slime73:
I haven't forgotten about this, I just need to think about it a bit but haven't had the time lately

MikuAuahDark:
No worries. Take your time.

slime73:
Let's try it out and see how it goes.

However we'll probably need to keep code around forever to do the slash replacement, to make sure it errors (or shows the deprecation notice) rather than not finding the file and therefore falling back to Lua's built in searcher.

PR #1791, opened 2022-03-29
ENet: Added variants for peer:send and host:broadcast which accepts love::Data

EngineerSmith:
Added variants for `peer:send` and `host:broadcast` which accepts `love::Data`. I got annoyed at having to use lua strings for enet.

I can understand if this pull request is rejected due to mixing love types into the libraries directory. It would be better to make a module, but after looking through the code base I think it's above me right now (and should be discussed if a network module is something love should have anyway). I do think it could benefit from being a module, as then events could be userdata with a `love::Data` field instead of a lua string integrating enet further into love. 

Working example:
```lua
-- setup client
local enet = require("enet")
local host = enet.host_create()
local server = host:connect("localhost:10001")
host:service(100)

-- populate byteData
local byteData = love.data.newByteData(6)
local ffi = require("ffi")
local ptr = ffi.cast("void*", byteData:getFFIPointer())
ffi.copy(ptr, "hello")

server:send(byteData) -- send byteData

love.update = function()
  host:service(10)
end
```
```lua
local enet = require("enet")
local host = enet.host_create("*:10001")

local text 
love.update = function()
  local event = host:service(10)
  if event then
    text = event.data
  end
end

love.draw = function()
  love.graphics.print(tostring(text)) -- prints 'hello' once client joins
end
```


slime73 (changes requested):
Maybe it makes sense to accept a lightuserdata + size? That would remove the dependency on love while still allowing you to pass a Data object's contents into enet (via `Data:getPointer` and `Data:getSize`).

EngineerSmith:
Okay, this didn't work as I thought it did - I'm going to close this and reopen a new PR with the suggested changes

PR #1804, opened 2022-05-01
Seamless Audio Device Switching

### LINKED ISSUE #1537: Selecting audio output
**[Original report](https://bitbucket.org/rude/love/issue/1532) by Snusmumriken (Bitbucket: [MainTomato](https://bitbucket.org/MainTomato), GitHub: [MainTomato](https://github.com/MainTomato)).**

Love at starting selects the default audio output device and it cannot be changed. It can also lead to errors when connecting USB headphones with their own sound card: when default audio device is speakers, and you connect USB-headphones, the sound continues to flow through the speakers. And when headphones is default and you turn them off, the sound stops working. This feature may solve this weird behavior.

MikuAuahDark:
Fixes #1537 

This requires OpenAL-soft version 1.22.0 or later, which is the case for our provided builds.

### DIFF
--- a/src/modules/audio/openal/DeviceWatcher.cpp
+++ b/src/modules/audio/openal/DeviceWatcher.cpp
@@ -1,0 +1,143 @@
+/**
+ * Copyright (c) 2006-2022 LOVE Development Team
+ *
+ * This software is provided 'as-is', without any express or implied
+ * warranty.  In no event will the authors be held liable for any damages
+ * arising from the use of this software.
+ *
+ * Permission is granted to anyone to use this software for any purpose,
+ * including commercial applications, and to alter it and redistribute it
+ * freely, subject to the following restrictions:
+ *
+ * 1. The origin of this software must not be misrepresented; you must not
+ *    claim that you wrote the original software. If you use this software
+ *    in a product, an acknowledgment in the product documentation would be
+ *    appreciated but is not required.
+ * 2. Altered source versions must be plainly marked as such, and must not be
+ *    misrepresented as being the original software.
+ * 3. This notice may not be removed or altered from any source distribution.
+ **/
+
+#include <chrono>
+
+// LOVE
+#include "common/delay.h"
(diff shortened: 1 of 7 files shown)


MikuAuahDark:
I personally don't like that it tries to reconnect through main thread in `love.run`.

Can `alcReopenDeviceSOFT` be called from any thread (or other than thread that's used for creating the original device)? @kcat

kcat:
> I personally don't like that it tries to reconnect through main thread in `love.run`.
> 
> Can `alcReopenDeviceSOFT` be called from any thread (or other than thread that's used for creating the original `ALCdevice`)? @kcat

As long as it's a thread that's otherwise safe to make AL calls on, to open/close/reset devices, yes. Places like `DllMain` on Windows or signal handlers might not be safe, because those are run in a special context by the system that may cause trouble with device communication, but normal threads that aren't doing anything special should be fine.

MikuAuahDark:
2nd iteration by using LOVE threads.

It still using C++'s high resolution timer though.

src/modules/audio/openal/DeviceWatcher.cpp
```diff
@@ -0,0 +1,143 @@
...
+DeviceWatcher::~DeviceWatcher()
+{
+}
+
+void DeviceWatcher::stop()
+{
+	stopped = true;
+}
+
+bool DeviceWatcher::tryReconnect(const ALCchar* deviceName)
```
slime73, line 68: I think I'd rather all actual reconnecting work only happen on the main thread, and not in the pool thread or the device watcher thread. The device watcher thread can continue to check for changes, which the main thread can query (in something called by `love.event.pump`, maybe).

Having device changes happen on multiple / different threads increases the chances of subtle race condition bugs a lot, and makes them harder to track down, compared to keeping it all in one place on the main thread.

MikuAuahDark:
After further discussion in #1537, I think I want to re-do the PR better.

PR #1805, opened 2022-05-03
Support for Audio Device Selection and Disconnection Callback

### LINKED ISSUE #1537: Selecting audio output

MikuAuahDark:
A simpler approach to #1804. Fixes #1537 _completely_.

This pull request adds several features:
* `love.audio.getPlaybackDevice()` function get current playback device name.
* `love.audio.getPlaybackDevices()` function retrieves all available playback devices in a Lua table.
* `love.audio.setPlaybackDevice(name)` function re-set the output device to `name`. `name` can be `nil` (for automatic default) or a string returned from `love.audio.getPlaybackDevices()` function above. Returns `true` on success or `false` and error message on failure.
* `love.audiodisconnected` callback when the audio device is disconnected and must be reconnected with `love.audio.setPlaybackDevice()`.

Now here are some notes:
* `love.audio.setPlaybackDevice()` is only supported if [ALC_SOFT_reopen_device](https://openal-soft.org/openal-extensions/SOFT_reopen_device.txt) extension is supported. Since you can't set to alternative device without this function, `love.audio.getPlaybackDevices()` is not really useful in that case. It could be possible to allow complete audio reinitialization with `love.audio.setPlaybackDevice()` before any other calls to `love.audio` is made to support scenario where "ALC_SOFT_reopen_device" extension is not supported but this adds slight complexity to the `Audio` class.
* `love.audiodisconnected` callback support depends on [ALC_EXT_disconnect](http://icculus.org/alextreg/wiki/action=printer&id=ALC_EXT_Disconnect) extension and is backend-dependent. It may or may not fire the callback when the audio is disconnected.
* When `love.audiodisconnected` callback is called, all previously playing sources are paused, then stopped a second later (`Source:tell()` returns 0). The latter may OpenAL-soft specific behavior though.
* Re-setting output device clears all the OpenAL internal buffer. ~~For "stream" sources, this desynchronize `Source:tell()` between the actual played audio, either slightly or significantly. For "static" sources, the source is stopped completely and must be re-played from the beginning.~~ I decided to stop all sources instead.

### DIFF
--- a/src/modules/audio/Audio.cpp
+++ b/src/modules/audio/Audio.cpp
@@ -78,7 +78,7 @@ bool Audio::setMixWithSystem(bool mix)
 #endif
 }
 
-void Audio::setOutputDevice(const char */*name*/)
+void Audio::setPlaybackDevice(const char */*name*/)
 {
 	throw love::Exception("Re-setting output device is not supported.");
 }
--- a/src/modules/audio/Audio.cpp
+++ b/src/modules/audio/Audio.cpp
@@ -78,6 +78,11 @@ bool Audio::setMixWithSystem(bool mix)
 #endif
 }
 
+void Audio::setOutputDevice(const char */*name*/)
+{
+	throw love::Exception("Re-setting output device is not supported.");
+}
+
 StringMap<Audio::DistanceModel, Audio::DISTANCE_MAX_ENUM>::Entry Audio::distanceModelEntries[] =
 {
 	{"none", Audio::DISTANCE_NONE},
--- a/src/modules/audio/Audio.h
+++ b/src/modules/audio/Audio.h
@@ -298,19 +298,19 @@ public:
 	virtual void resumeContext() = 0;
 
 	/**
-	 * Get current output device name.
+	 * Get current playback device name.
 	 */
-	virtual std::string getOutputDevice() = 0;
+	virtual std::string getPlaybackDevice() = 0;
 
 	/**
-	 * Retrieve list of available output devices.
+	 * Retrieve list of available playback devices.
 	 */
-	virtual void getOutputDevices(std::vector<std::string> &list) = 0;
+	virtual void getPlaybackDevices(std::vector<std::string> &list) = 0;
 
 	/**
-	 * Set the output device to specified device name.
+	 * Set the current playback device to specified device name.
 	 */
-	virtual void setOutputDevice(const char *name);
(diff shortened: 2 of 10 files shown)


MikuAuahDark:
There are something that I want to implement. It's possible to retrieve all the source position before calling `Source::stop(pool)`.
* Should this be implemented and all the position of the sources are passed to `love.audiodisconnected` callback?
* If so, what would be the units? Seconds or samples?

slime73:
> There are something that I want to implement. It's possible to retrieve all the source position before calling Source::stop(pool).
> Should this be implemented and all the position of the sources are passed to love.audiodisconnected callback?

Would it be possible to automatically seek the sources to their last playback position after stopping them (and before passing them to the callback)? Or would that need OpenAL state that isn't working by that point?

I also wonder if there needs to be any special consideration for queueable sources when doing that (e.g. not seeking those).

MikuAuahDark:
Not possible. The audio state is disconnected during the `love.audiodisconnected` callback, so no `:play` or `:seek` should be called unless `love.audio.setPlaybackDevice` is called (and the user must return `true` so the callback didn't reconnect twice). The `love.audiodisconnected` callback reconnects the device to default device if user returns false or nil. I think this could be changed to check for string so that it reconnects to specific device.

slime73:
> Not possible. The audio state is disconnected during the love.audiodisconnected callback

The tell and seek implementations have conditional codepaths depending on whether there's a valid OpenAL source associated with that love Source, though. So as long as they know that there's no valid OpenAL source after the disconnect, tell/seek should work, right?

MikuAuahDark:
I tested it real quick and while it works for stream sources, it doesn't work for static sources at all. `alGetSourcei(source, AL_SAMPLE_OFFSET, &offset)` in `Source::tell` returns 0 in the `offset`.

slime73:
Ok, let's put this in now and we can figure out possible streamlining and/or name changes later

PR #1829, opened 2022-08-11
fix cmake

Er2ch:


CMakeLists.txt
```diff
@@ -1073,6 +1073,27 @@ set(LOVE_SRC_MODULE_WINDOW
...
+# Love scripts
+# Useful while developing
+#
+
+set(LOVE_SRC_SCRIPTS
+    src/scripts/nogame.lua
+)
+
+find_program(LOVE_LUA_PROGRAM NAMES lua luajit lua54 lua53 lua52 lua51)
+if(LOVE_LUA_PROGRAM)
```
slime73, line 1086: I plan to eventually delete auto.lua if we can find a way to stop the no-game code from needing it (similar to all the other in-tree lua files in love's code, which no longer need auto.lua). So I'm not sure about adding stuff like this and the auto.lua changes

Er2ch:
Removed auto.lua building (I also haven't got idea how to remove it except using raw strings inside) and fixed GCC compilation error.

readme.md
```diff
@@ -53,6 +53,11 @@ Run `platform/unix/automagic` from the repository root, then run ./configure and
 
 When using a source release, automagic has already been run, and the first step can be skipped.
 
+Also possible to use CMake build platform.
+
+	$ cmake -B Release/ -H. -G "Unix Makefiles"
```
MikuAuahDark, line 58: Consider `-S.` instead of `-H.`.

MikuAuahDark:
Speaking of the this PR, I don't think the Linux OS workflow uses the CMake, so it's probably a good idea to have another Linux workflow that uses the CMake. It doesn't have to create AppImage of course, at least the compile must successful (and optionally run `love --version` without problems).

changes.txt
```diff
@@ -46,6 +46,7 @@ Released: N/A
 * Added love.system.getPreferredLocales.
 * Added love.localechanged callback.
 * Added World:getFixturesInArea().
+* Added CMake build ability.
```
slime73, line 49: Windows CMake has always worked and continues to work. I don't think we really want to officially support multiple build paths on each platform (although unofficial support is fine), so I'd rather not have this in the changelog.

readme.md
```diff
@@ -53,6 +53,11 @@ Run `platform/unix/automagic` from the repository root, then run ./configure and
 
 When using a source release, automagic has already been run, and the first step can be skipped.
 
+Also possible to use CMake build platform.
```
slime73, line 56:
```suggestion
CMake may also work, although it is not the primary recommended build path.
```

slime73:
The main changes here were done separately in other commits (mostly https://github.com/love2d/love/commit/51e3cf2a9c8a49f27f2467d2b95ad19bdf9ff70f), so I'll close this.

PR #1831, opened 2022-08-20
SDL 2.24.0 support

Er2ch:


src/modules/joystick/Joystick.cpp
```diff
@@ -58,16 +58,20 @@ STRINGMAP_CLASS_END(Joystick, Joystick::Hat, Joystick::HAT_MAX_ENUM, hat)
...
+	{ "xboxone",      Joystick::GAMEPAD_TYPE_XBOXONE             },
+	{ "ps3",          Joystick::GAMEPAD_TYPE_PS3                 },
+	{ "ps4",          Joystick::GAMEPAD_TYPE_PS4                 },
+	{ "ps5",          Joystick::GAMEPAD_TYPE_PS5                 },
+	{ "switchpro",    Joystick::GAMEPAD_TYPE_NINTENDO_SWITCH_PRO },
+	{ "amazonluna",   Joystick::GAMEPAD_TYPE_AMAZON_LUNA         },
+	{ "stadia",       Joystick::GAMEPAD_TYPE_STADIA              },
+	{ "virtual",      Joystick::GAMEPAD_TYPE_VIRTUAL             },
+	{ "shield",       Joystick::GAMEPAD_TYPE_NVIDIA_SHIELD       },
+	{ "joycon_left",  Joystick::GAMEPAD_TYPE_JOYCON_LEFT         },
```
slime73, line 72: These should be `"joyconleft"`, `"joyconright"`, and `"joyconpair"` to match love's existing enum string naming scheme.

src/modules/joystick/sdl/Joystick.cpp
```diff
@@ -509,7 +515,7 @@ bool Joystick::setVibration(float left, float right, float duration)
 	Uint32 length = SDL_HAPTIC_INFINITY;
 	if (duration >= 0.0f)
 	{
-		float maxduration = std::numeric_limits<Uint32>::max() / 1000.0f;
+		float maxduration = std::numeric_limits<Uint32>::max() / 1000;
```
slime73, line 518: An integer divided by an integer will implicitly do integer division, which isn't what we want here I think. To avoid the precision issues with converting uint32_max to a float, converting it to a double should work I believe:

```suggestion
		float maxduration = (float) (std::numeric_limits<Uint32>::max() / 1000.0);
```

slime73 (changes requested):

Er2ch:
Changes resolved.

src/modules/joystick/sdl/Joystick.cpp
```diff
@@ -509,7 +515,7 @@ bool Joystick::setVibration(float left, float right, float duration)
 	Uint32 length = SDL_HAPTIC_INFINITY;
 	if (duration >= 0.0f)
 	{
-		float maxduration = std::numeric_limits<Uint32>::max() / 1000.0f;
+		float maxduration = (float)std::numeric_limits<Uint32>::max() / 1000.0;
```
slime73, line 518: Won't this still cast `2^32-1` to a float, and then do the division? That was the original issue with the code that you wanted to change, because floats don't have enough precision to represent 2^32-1 (it ends up as 2^32 instead).

I believe adding parentheses around it would resolve it because I think it will cast 2^32-1 to a double instead of a float, for the division:
```suggestion
		float maxduration = (float) (std::numeric_limits<Uint32>::max() / 1000.0);
```

Issue #1760, opened 2022-01-03
Unexpected behavior of love.graphics.line with duplicate points


megagrump:
```lua
function love.draw()
	love.graphics.line(0, 0, 100, 100, 100, 100) -- a
	love.graphics.line(0, 0, 50, 50, 50, 50, 100, 100) --b
	love.graphics.line(0, 0, 50, 50, 100, 100, 50, 50) -- c
end
```
I would expect all of these calls to draw a line from `0, 0` to `100, 100`, but neither of them do.

a and b draw nothing. c draws a line from `0, 0` to `50, 50`.

```lua
function love.draw()
	love.graphics.line(100, 100, 100, 100)
end
```
Since line coordinates are endpoint-inclusive, I think this should draw a single point at `100, 100`.

Funny backstory: I wasted a considerable amount of time trying to find a bug in a routine that builds a huge graph; turns out all the nodes were actually connected, but some of the lines in the visualisation had duplicate points like this.

slime73:
For reference, there's a bit of discussion about this in #972. However I haven't decided yet if the same stance should be taken as in the original issue so I might leave this open.

megagrump:
Hrm, I should have searched more thoroughly.

At the very least document the exact behavior and put it in one of those orange boxes in the wiki.

It is very surprising to me that

```lua
function love.draw()
	love.graphics.line(0, 0, 100, 100, 0, 0)
end
```
and
```lua
function love.draw()
	love.graphics.line(0, 0, 100, 100)
	love.graphics.line(100, 100, 0, 0)
end
```
give different results.

radgeRayden:
I think especially when we're talking about arbitrarily generated lines, it would be very nice to have a more resilient line drawing function, and in particular the expected behaviour of drawing a point if start and end are the same simplifies things a bit (for the end user). Maybe it is unexpected for other people, though.

slime73:
pgimeno did a bit of investigation into the problem, here: https://love2d.org/forums/viewtopic.php?f=3&t=92555

slime73:
This is fixed as of [6b54ee5](https://github.com/love2d/love/commit/6b54ee50db121f907c38dfb882970f2826b12853) (part of love 12, although we could backport the fix I suppose if we end up releasing a new 11.x version).

PR #1842, opened 2022-09-16
FIx polyline adding NaN normals, which breaks the line rendering

Klonan:
Resolves https://github.com/love2d/love/issues/1760

The drawing was having these NAN normals inserted, because the segment length was 0.
This adds asserts to check the normals are never NAN, and early return to skip the line segment if the length is 0.

-------------------------------------------

Before, using this in Love2d:
```lua
love.graphics.line(0, 0, 50, 50, 50, 50, 100, 100)
```

Results in the line draw normals looking like this:

Code_I2fkkV4mG9

Which through some mechanism, doesn't crash, but doesn't draw the line either (points drawn for debug):

image

With the skipping of length 0 segments, the normals will look valid, and the line will draw as expected:

Code_nWykJnNO3R

image

-----------------------------------

Duplicate points don't seem to cause any problem with the "none" and "bevel" join styles, so I didn't worry about skipping the points for those cases (though it would probably be fine and save some wasted time)

### DIFF
--- a/src/modules/graphics/Polyline.cpp
+++ b/src/modules/graphics/Polyline.cpp
@@ -24,6 +24,7 @@
 
 // C++
 #include <algorithm>
+#include <assert.h>
 
 // treat adjacent segments with angles between their directions <5 degree as straight
 static const float LINES_PARALLEL_EPS = 0.05f;
@@ -176,6 +177,12 @@ void MiterJoinPolyline::renderEdge(std::vector<Vector2> &anchors, std::vector<Ve
 {
 	Vector2 t    = (r - q);
 	float len_t = t.getLength();
+	if (len_t == 0.0f)
+	{
+		// degenerate segment, skip it
+		return;
+	}
+
 	Vector2 nt   = t.getNormal(hw / len_t);
 
 	anchors.push_back(q);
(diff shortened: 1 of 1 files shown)


src/modules/graphics/Polyline.cpp
```diff
@@ -185,6 +192,7 @@ void MiterJoinPolyline::renderEdge(std::vector<Vector2> &anchors, std::vector<Ve
 	if (fabs(det) / (len_s * len_t) < LINES_PARALLEL_EPS && Vector2::dot(s, t) > 0)
 	{
 		// lines parallel, compute as u1 = q + ns * w/2, u2 = q - ns * w/2
+		assert(ns == ns); //NaN check
```
slime73, line 195: We don't generally use C's `assert` in love's code because it's not too helpful for users since it won't get converted to a Lua error (whereas throwing a `love::Exception` will).

Should this throw an exception in that case? What user input will cause NaNs here, given the degenerate segment check above?

Klonan: The asserts are more for proving that the fix works, in that you can remove the changed lines, and observe that the assert was hit.

For now there would be no other way for the NaNs so sneak in, but if someone changed/reorginized/refactored the surrounding code in the future, the exceptions are a reminder that this issue is possible.

As well, if there was a testSuite, then the exceptions are a nice direct way of testing for the functionality, its just 'Run the code' and the asserts will terminate, instead of someway testing that the line segments were drawn with a specific set of points.

-------------------------------------

I can change them to love:Exception if you'd like, or remove them if you prefer

PR #1844, opened 2022-09-20
Remove dependency of system vulkan headers

nikeinikei:
Installing the vulkan sdk is now optional, if you need the validation layers.


slime73:
nice, thanks!

PR #1850, opened 2022-09-24
Add optional load mode parameter to love.filesystem.load.

### LINKED ISSUE #1849: Add optional parameter to love.filesystem.load to prevent loading bytecode
LuaJIT supports Lua 5.2+'s APIs for this via `luaL_loadbufferx`. The parameter should probably match stock Lua's: https://www.lua.org/manual/5.3/manual.html#pdf-load

MikuAuahDark:
This adds optional `loadMode` parameter to `love.filesystem.load(path, loadMode)`. `loadMode` can be `"b"` to only accept binary chunk, `"t"` to only accept text chunk, or `"bt"` for both. `"bt"` is the default behavior if not specified. Fixes #1849

Note that this effectively prevents people who trying to switch between Lua 5.1 and LuaJIT in Windows by replacing the DLL. If this is undesired please let me know so I can think of another approach.

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -9,6 +9,7 @@ Released: N/A
 * Added love.filesystem.mountFullPath and love.filesystem.unmountFullPath, including opt-in mount-for-write support.
 * Added love.filesystem.mountCommonPath, unmountCommonPath, and getFullCommonPath.
 * Added 'readonly' field to love.filesystem.getInfo's returned table.
+* Added an optional load mode parameter to love.filesystem.load whetever to only allow binary chunks, text chunks, or both.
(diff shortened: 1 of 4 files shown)


slime73:
> Note that this effectively prevents people who trying to switch between Lua 5.1 and LuaJIT in Windows by replacing the DLL. If this is undesired please let me know so I can think of another approach.

While this is a bit unfortunate, I wonder how many people / use cases this actually affects in practice. Maybe we can put it in like this now and see if anyone runs into it while testing 12.0, since it might get pretty annoying to make it work more dynamically.

Issue #1002, opened 2015-02-22
Noise Function's Implementation may be Patented


slime73:
**[Original report](https://bitbucket.org/rude/love/issue/997) by Zauber Paracelsus (Bitbucket: [zauberparacelsus](https://bitbucket.org/zauberparacelsus), GitHub: [zauberparacelsus](https://github.com/zauberparacelsus)).**

----------------------------------------

The love.math.noise function is an implementation of the Simplex noise algorithm, and generates noise up to 4 dimensions.  The problem, however, is that Simplex Noise unfortunately is covered by a patent for 3D and higher implementations.

Is Love using the patented implementation?  Or is it using an implementation unaffected by the patent, like what OpenSimplex did?

If Love is using the patented implementation, then perhaps alternatives need to be looked at.  There is the aforementioned OpenSimplex implementation, though it is somewhat slower than Simplex (not sure how much).  There is also the Wavelet Noise implementation made by Pixar, which like Simplex avoids the artefacts found in Perlin noise.

OpenSimplex: https://github.com/smcameron/open-simplex-noise-in-c
Wavelet Noise: http://graphics.pixar.com/library/WaveletNoise/paper.pdf

Footnote 1: Although I set the issue type to Bug, this is more of a general issue than a bug, and I didn't think the other issue types were any more appropriate.

Footnote 2: There are at least a one or two confused claims scattered online that the patent expired in 2005.  This is untrue, and 2005 was actually the year that it was published.

PR #1851, opened 2022-09-24
Separate love.math.noise to perlinNoise and simplexNoise

MikuAuahDark:
As per Discord discussion, patent of 3D and 4D simplex noise is no longer a concern (#1002). This pull request adds 2 functions and deprecate 1 function.
* `love.math.perlinNoise(x[, y[, z[, w]]])` which generates 1D-4D Perlin Noise
* `love.math.simplexNoise(x[, y[, z[, w]]])` which generates 1D-4D Simplex Noise
* Deprecate `love.math.noise` to prevent confusion.

There was a few ideas on how to specify the noise function, but I found out separating it to 2 functions is the best approach in terms of speed. The other proposal was to use `love.math.noise(noiseType, x[, y[, z[, w]]])` but this has some performance overhead on non-JIT-ed environments.

### DIFF
--- a/src/libraries/noise1234/noise1234.cpp
+++ b/src/libraries/noise1234/noise1234.cpp
@@ -11,6 +11,8 @@
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 // General Public License for more details.
 
+// Modified by the LOVE Development Team to use double precision.
+
 /** \file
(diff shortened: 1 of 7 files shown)


slime73:
I'm open to changing the API between now and 12.0's release, although I still lean towards this style compared to alternatives proposed elsewhere.

PR #1854, opened 2022-10-02
Add Data:getIntArray

IoriBranford:
# Data:getIntArray

A convenient shortcut for the common case of reading data as an int array.

```lua
--Before
    local array = {}
    local i, n = 1, data:getSize()
    while i <= n do
        array[#array + 1], i = love.data.unpack("<I4", data, i)
    end

--After
    array = data:getIntArray('u', 4, 'l')
```

## Arguments

- string `signedness`
  - 'u' for unsigned. Anything else for signed. Default is signed.
- number `intsize`
  - How many bytes in each int. Supports 1, 2, 4, or 8. Default is 4. Any other value causes an error.
- string `endianness`
  - Which byte order the ints are stored in. 'l' for little endian, 'b' for big endian. Default is the running machine's endianness.

## Returns

- table `intarray`
  - Table of the numbers read from the data.

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -48,6 +48,7 @@ Released: N/A
 * Added love.system.getPreferredLocales.
 * Added love.localechanged callback.
 * Added World:getFixturesInArea().
+* Added Data:getIntArray().
 
 * Changed the default font from Vera size 12 to Noto Sans size 13.
 * Changed the Texture class and implementation to no longer have separate Canvas and Image subclasses.
--- a/src/modules/data/wrap_Data.cpp
+++ b/src/modules/data/wrap_Data.cpp
@@ -42,6 +42,68 @@ int w_Data_getString(lua_State *L)
 	return 1;
 }
 
+template <class Int> void getIntArray(lua_State *L, size_t size, void *data)
+{
+	Int *idata = (Int *) data;
+	size_t length = size / sizeof(Int);
+	lua_createtable(L, (int)length, 0);
+	for (size_t i = 1; i <= length; ++i)
+	{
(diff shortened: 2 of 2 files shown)


slime73:
On desktop and mobile platforms you can also use the FFI to access the data directly as a uint array. Something like this:
```Lua
local ffi = require("ffi")

local uintarray = ffi.cast("uint32_t *", data:getFFIPointer())

print(tonumber(uintarray[0])) -- FFI C arrays are 0-indexed.
```

What are you using this for? I've thought about adding some sort of get/set APIs to ByteData (not arbitrary Data), but designing something that's useful and general without being too slow or cumbersome to use is a bit challenging. I'm not sure just uint arrays will necessarily cover many use cases, but I don't really know what the use cases are in general.

IoriBranford:
Game map files can have binary int arrays in some form, for example [Tiled](https://github.com/mapeditor/tiled) maps with compressed layers. It writes the layer contents as base64 strings of compressed uint32 arrays (little endian, which reminds me I need to add an endianness option).
It would use this function like:
```lua
    if encoding == "base64" then
        data = love.data.decode("data", encoding, data)
        if compression then
            data = love.data.decompress("data", compression, data)
        end
    end

    return data:getIntArray('u', 4)
```

slime73:
I ended up adding Data methods to get integers and floats of different sizes at a given byte offset in a Data object. I think that covers this type of situation - I'm usually kind of hesitant to return arrays of things because they generate garbage and aren't too flexible (especially if you don't actually need an array), although the APIs I added to let you specify a number of elements to return as separate return values.

PR #1855, opened 2022-10-05
vulkan: fix incorrect uniform values

nikeinikei:


src/modules/graphics/vulkan/StreamBuffer.cpp
```diff
@@ -95,13 +95,13 @@ love::graphics::StreamBuffer::MapInfo StreamBuffer::map(size_t /*minsize*/)
 
 	MapInfo info;
 	info.size = bufferSize - frameGPUReadOffset;
-	info.data = (uint8*)allocInfo.pMappedData + (frameIndex * bufferSize) + frameGPUReadOffset;
+	info.data = (uint8*)allocInfo.pMappedData + (vgfx->getFrameIndex() * bufferSize) + frameGPUReadOffset;
```
slime73, line 98: Is this part necessary for the fix or just code cleanup? I kind of like having a StreamBuffer's frame counter local to itself, in case we want to change how things work just for StreamBuffers (for example the metal backend has 3 frames worth of space in its StreamBuffers, and the OpenGL backend has 4).

nikeinikei: I just tested and it is not necessary for the fix. I forgot to remove that change again with all the different changes I tried for debugging. I'll revert it.

PR #1856, opened 2022-10-08
Update LuaSocket to 3.1.0

### LINKED ISSUE #1790: LuaSocket has released version 3.0.0
https://github.com/lunarmodules/luasocket/releases/tag/v3.0.0

Do you think we should put it in Love 12.x?

MikuAuahDark:
This PR updates LuaSocket to 3.1.0, along with minor refactor on how LuaSocket is preloaded and support for Unix stream socket under Windows. The LuaSocket version writes 3.0.0 but looks like they [forgot to change it](https://github.com/lunarmodules/luasocket/blob/v3.1.0/src/luasocket.h#L13).

**Note: A manual change for the macOS project is required. Since I don't have macOS, I can't do this part.**

Fixes #1790


MikuAuahDark:
Weird. Looks like downloads to CMake is blocked for linux-os workflow, but nonetheless it compiles fine in the individual commit.

Issue #1735, opened 2021-10-15
Support -h/--help cli flags


gvanem:
I first heard of Love2D today.
Downloading the Win release .zip and trying it required some effort (no real instructions in the `readme.txt`).
For example:
```
lovec.exe -h
Error: [string "boot.lua"]:577: Cannot load game at path 'F:/MingW32/src/Games/Love2D/-h'.
Make sure a folder exists at the specified path.
...
```

(same with `love.exe -h`). 

Reading the Wiki, copy & paste the sample to `hello-world.lua` and running `lovec.exe hello-world.lua` 
gave me same bleeding stupid Lua call-stack. 

Then voila renaming `hello-world.lua` to  `main.lua` and running `lovec.exe .` gave what I'd expect.

How hard should it be?  Please handle a `-h` / `--help` option like any program with some respect for itself would.

slime73:
It might make sense to print a stripped down version of what's in the [Getting Started](https://love2d.org/wiki/Getting_Started) wiki page, when --help is passed in.

However, since you apparently already read the wiki (where that page says to name your file main.lua and pass in the folder containing the file), I'm not sure it would have helped you.

We'd also need to be careful to avoid preventing games from implementing their own help arguments, if that's desired.

marcotrosi:
I agree, but confusing title for this issue. "Support -h/--help cli flags"

gvanem:
> However, since you apparently already read the wiki 

I read this: https://love2d.org/wiki/Main_Page which says:
 *Running this code will cause an 800 by 600 window to appear, and display white text on a black background.*

It does not mention how. Hence my confusion since I only skimmed some other Wiki pages.

hahawoo:
Running `love.exe hello-world.lua` intuitively seems like it should work, and not only does it not work but there's no way to learn from the command line what to do.

(I think that this particular instance of confusion could have been prevented with improved documentation, but my opinions on documentation are beyond the scope of this issue.)

That said, I'm not sure about this proposal.

When would this help message be shown? Only when `-h/--help` is passed? What about `love.exe hello-world.lua`? If so, this would mean that LOVE project directories couldn't end in ".lua" (or be "-h" for that matter). What about when just `love.exe` is run with no arguments? I guess this would still have to run the no-game screen as well, since love.exe could also be run from the OS GUI.

[...]

PR #1857, opened 2022-10-20
print usage when there is no game and when --help option is used

premek:
Hi, I wanted to prepare a PR for #1735, does something like this look ok?

I added a `--help` option in the same way as there is `--version` now. It's used only when it's before a game dir or a love file, otherwise it's passed to the game arguments, example:

```
love --help #prints usage
love path/to/game --help #passed to the game
love game.love --help #passed to the game
```

I also printed usage when there is "no game" displayed, is that ok?

Could someone please check my English in the usage message (or propose a completely different wording?)

Is there a place where to put the string and use it on both places (see the diff)?

Thanks

### DIFF
--- a/src/love.cpp
+++ b/src/love.cpp
@@ -141,6 +141,21 @@ enum DoneAction
 	DONE_RESTART,
 };
 
+static void print_usage()
+{
+    // when editing this message, change it at boot.lua too
+    printf("LÖVE is an *awesome* framework you can use to make 2D games in Lua\n"
+        "https://love2d.org\n"
+        "\n"
+        "usage:\n"
+        "    love --version                  prints LÖVE version and quits\n"
+        "    love --help                     prints this message and quits\n"
+        "    love path/to/gamedir            runs the game from the given directory which contains a main.lua file\n"
+        "    love path/to/packagedgame.love  runs the packaged game from the provided .love file\n"
+        "    love path/to/file.lua           runs the game from the given .lua file\n"
+        );
+}
+
 static DoneAction runlove(int argc, char **argv, int &retval, love::Variant &restartvalue)
 {
 	// Oh, you just want the version? Okay!
@@ -151,6 +166,13 @@ static DoneAction runlove(int argc, char **argv, int &retval, love::Variant &res
 		love_openConsole(err);
 #endif
 		printf("LOVE %s (%s)\n", love_version(), love_codename());
+		retval = 0;
+		return DONE_QUIT;
+	}
+
+	if (argc > 1 && strcmp(argv[1], "--help") == 0)
+	{
+		print_usage();
 		retval = 0;
 		return DONE_QUIT;
 	}
--- a/src/modules/love/boot.lua
+++ b/src/modules/love/boot.lua
@@ -135,6 +135,17 @@ function love.boot()
 	end
 
 	if not can_has_game then
+        -- when editing this message, change it at love.cpp too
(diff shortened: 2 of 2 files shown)


src/love.cpp
```diff
@@ -141,6 +141,17 @@ enum DoneAction
...
 };
 
+static void print_usage()
+{
+	printf("usage:\n"
+		"    love --version                  prints LÖVE version and quits\n"
+		"    love --help                     prints this message and quits\n"
+		"    love path/to/gamedir            runs the game from the given directory which contains main.lua file\n"
+		"    love path/to/packagedgame.love  runs the packaged game from the provided .love file\n"
+		"    love path/to/main.lua           runs the game from the given .lua file\n"
```
slime73, line 151: Maybe this example could use a different filename than `main.lua`, so it's more clear that the name doesn't need to be that:
```suggestion
		"    love path/to/file.lua           runs the game from the given .lua file\n"
```

src/love.cpp
```diff
@@ -141,6 +141,17 @@ enum DoneAction
 	DONE_RESTART,
 };
 
+static void print_usage()
+{
+	printf("usage:\n"
+		"    love --version                  prints LÖVE version and quits\n"
+		"    love --help                     prints this message and quits\n"
+		"    love path/to/gamedir            runs the game from the given directory which contains main.lua file\n"
```
slime73, line 149:
```suggestion
		"    love path/to/gamedir            runs the game from the given directory which contains a main.lua file\n"
```

src/love.cpp
```diff
@@ -141,6 +141,17 @@ enum DoneAction
 	DONE_RESTART,
 };
 
+static void print_usage()
+{
+	printf("usage:\n"
```
slime73, line 146: I wonder if it makes sense to link to the homepage or wiki, after the usage examples?

src/love.cpp
```diff
@@ -141,6 +141,17 @@ enum DoneAction
...
+		"    love --version                  prints LÖVE version and quits\n"
```
slime73, line 147: Maybe each description sentence/line should end with a period (`.`) - although I don't know if it's common to not do that for this type of thing.

slime73 (changes requested):
The change looks good - there's just a few things that could be cleaned up a bit in the print message.

src/love.cpp
premek: I looked at a few linux commands' usage and I think none of them had the period at the end of each line. Some had capital first letter, most had lowercase.

src/love.cpp
premek: Added one sentence from the homepage and a link

src/love.cpp
premek: Fixed, thanks you

premek:
Hi, thanks a lot for looking at this, I did the changes (see my comments). Is it ok that the same message would be copied in 2 files?

I thought maybe if it was only in love.cpp and displayed for `--help` and when `argc==0`?

slime73:
> Is it ok that the same message would be copied in 2 files?

Yeah I think it's probably fine, but we'll try it and see how it feels to have the print when the no-game screen is loaded.

slime73 (approved):

MikuAuahDark:
A berrer way to deal with the duplicate usage text would be having the `print_usage` function exposed to the Lua side (e.g. `love._printUsageHelp()`) then the `boot.lua` can call that function instead.

slime73:
That's definitely a possibility, I'm not sure if it's worth it for such a small thing though.

PR #1861, opened 2022-11-02
Fix #1834: Give MouseJoint a sensible stiffness and damping initial value

### LINKED ISSUE #1834: [12.0] Mouse Joint not working
Might be related to the Box2D update from 2.3 to 2.4.1.
It is a plain default mouseJoint

11.4:

https://user-images.githubusercontent.com/11986037/186956659-38f50ddd-7345-462b-a746-deb2c8372a44.mp4

12.0:

https://user-images.githubusercontent.com/11986037/186956772-0e9df7ef-e5b8-41db-8f58-9f1618b14dde.mp4

zombrodo:
As it says on the tin, this is two commits to fix the behaviour in #1834. The first fixes it with minimal changes (by setting the default values from before), and the second is a draft of a more significant change

[Box2d 2.4.0 converted all its joints to use stiffness and damping](https://github.com/erincatto/box2d/blob/main/CHANGELOG.md#dynamics), rather than the old spring-damper, which meant that the springy-ness is defined by `stiffness`, rather than `frequency`. They provided a helper function for linear and angular stiffness to convert from the old way to the new way

When https://github.com/love2d/love/pull/1613 was merged, thegrb93 helpfully added in some conversion steps to continue taking the frequency/damping values, and under the hood use the helper functions to convert them to stiffness (as well as exposing getters and setters for `stiffness` and `damping`)

`love.physics` has (in my experience) been pretty close to the actual Box2d API, just with a few helpers. It's kinda helpful being so close, as it means you can often find materials talking about Box2d, and easily apply it to love. In that sense, rather than keeping both spring physics as functions on Joints, just expose conversions at the top `love.physics` level, and then use that to set values on the joints via setters:

```lua
local joint = love.physics.newMouseJoint(body, love.mouse.getPosition())
local stiffness, damping = love.physics.linearStiffness(5.0, 0.7, body)
joint:setStiffness(stiffness)
joint:setDamping(damping)
```

This is kinda nice, because it isolates the variation to one spot, rather than on each joint, so further API changes are minimal, I guess.

I'm more seeking feedback on the approach - I've marked it as a `Draft` because honestly? I'm not sold on it, so I'm keen for feedback (and happy to drop it if we want). On top of that, there are any C++ changes, lemme know ('cause this is the most I've written ehe)

### DIFF
--- a/src/modules/physics/box2d/Physics.cpp
+++ b/src/modules/physics/box2d/Physics.cpp
@@ -405,7 +405,12 @@ b2AABB Physics::scaleUp(const b2AABB &aabb)
 	return t;
 }
 
-void Physics::b2LinearFrequency(float& frequency, float& ratio, float stiffness, float damping, b2Body* bodyA, b2Body* bodyB)
+void Physics::computeLinearStiffness(float &stiffness, float &damping, float frequency, float dampingRatio, b2Body *bodyA, b2Body *bodyB)
+{
+	b2LinearStiffness(stiffness, damping, frequency, dampingRatio, bodyA, bodyB);
+}
+
+void Physics::computeLinearFrequency(float &frequency, float &ratio, float stiffness, float damping, b2Body *bodyA, b2Body *bodyB)
 {
 	float massA = bodyA->GetMass();
 	float massB = bodyB->GetMass();
@@ -435,7 +440,12 @@ void Physics::b2LinearFrequency(float& frequency, float& ratio, float stiffness,
 	ratio = damping / (mass * 2.0f * omega);
 }
 
-void Physics::b2AngularFrequency(float& frequency, float& ratio, float stiffness, float damping, b2Body* bodyA, b2Body* bodyB)
+void Physics::computeAngularStiffness(float &stiffness, float &damping, float frequency, float dampingRatio, b2Body *bodyA, b2Body *bodyB)
+{
+	b2AngularStiffness(stiffness, damping, frequency, dampingRatio, bodyA, bodyB);
+}
+
+void Physics::computeAngularFrequency(float &frequency, float &ratio, float stiffness, float damping, b2Body *bodyA, b2Body *bodyB)
 {
 	float IA = bodyA->GetInertia();
 	float IB = bodyB->GetInertia();
--- a/src/modules/physics/box2d/Physics.h
+++ b/src/modules/physics/box2d/Physics.h
@@ -360,18 +360,18 @@ public:
 	static b2AABB scaleUp(const b2AABB& aabb);
 
 	/**
-	 * Calculates linear frequency and damping radio from stiffness and damping
-	 * @param frequency The output frequency
-	 * @param ratio The output damping ratio
-	 * @param stiffness The joint stiffness
-	 * @param damping The joint damping
+	 * Calculates the stiffness and damping, given the linear frequency, and damping ratio.
+	 * @param stiffness The output stiffness
+	 * @param damping The output damping
+	 * @param frequency The joint linear frequency
+	 * @param dampingRatio The joint damping ratio
 	 * @param bodyA The bodyA of the joint
 	 * @param bodyB The bodyB of the joint
 	 **/
-	static void b2LinearFrequency(float& frequency, float& ratio, float stiffness, float damping, b2Body* bodyA, b2Body* bodyB);
+	static void computeLinearStiffness(float& stiffness, float& damping, float frequency, float dampingRatio, b2Body* bodyA, b2Body* bodyB);
 
 	/**
-	 * Calculates angular frequency and damping radio from stiffness and damping
+	 * Calculates linear frequency and damping ratio from stiffness and damping
 	 * @param frequency The output frequency
 	 * @param ratio The output damping ratio
 	 * @param stiffness The joint stiffness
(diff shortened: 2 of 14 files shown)


src/modules/physics/box2d/wrap_Physics.cpp
```diff
@@ -477,12 +477,100 @@ int w_setMeter(lua_State *L)
...
 
 }
+
 int w_getMeter(lua_State *L)
 {
 	lua_pushinteger(L, Physics::getMeter());
 	return 1;
 }
 
+int w_linearStiffness(lua_State* L) {
```
slime73, line 487: love's typical function naming convention is to have a verb at the start of the function name - in this case I think `compute` makes sense, e.g. `computeLinearStiffness`, etc.

That convention should be applied to the C++ code as well as the Lua APIs.

src/modules/physics/box2d/wrap_Physics.cpp
```diff
@@ -477,12 +477,100 @@ int w_setMeter(lua_State *L)
...
+	float frequency = (float)luaL_checknumber(L, 1);
+	float dampingRatio = (float)luaL_checknumber(L, 2);
+	Body* body1 = luax_checkbody(L, 3);
+	b2Body* other = 0;
+
+	if (lua_isnoneornil(L, 4)) {
```
slime73, line 493: love's code style has braces on separate lines - and for situations like this where there's only one line inside an `if`, the braces can be completely removed.

src/modules/physics/box2d/wrap_Physics.cpp
```diff
@@ -477,12 +477,100 @@ int w_setMeter(lua_State *L)
...
 int w_getMeter(lua_State *L)
 {
 	lua_pushinteger(L, Physics::getMeter());
 	return 1;
 }
 
+int w_linearStiffness(lua_State* L) {
+	float frequency = (float)luaL_checknumber(L, 1);
+	float dampingRatio = (float)luaL_checknumber(L, 2);
+	Body* body1 = luax_checkbody(L, 3);
```
slime73, line 490: love's code style has pointer and reference keywords (`*` and `&`) next to the variable name instead of the type name, e.g. `Body *body1`.

slime73 (commented):
I think this approach seems fairly sensible - there's just some naming and code style cleanup that should be done.

src/modules/physics/box2d/wrap_Physics.cpp
zombrodo: Ah, misconfigured auto-format! Fixed

src/modules/physics/box2d/Physics.cpp
```diff
@@ -405,7 +405,11 @@ b2AABB Physics::scaleUp(const b2AABB &aabb)
 	return t;
 }
 
-void Physics::b2LinearFrequency(float& frequency, float& ratio, float stiffness, float damping, b2Body* bodyA, b2Body* bodyB)
+void Physics::computeLinearStiffness(float &stiffness, float &damping, float frequency, float dampingRatio, b2Body *bodyA, b2Body *bodyB) {
```
slime73, line 408: Braces should be on a newline

src/modules/physics/box2d/Physics.cpp
```diff
@@ -435,7 +439,11 @@ void Physics::b2LinearFrequency(float& frequency, float& ratio, float stiffness,
 	ratio = damping / (mass * 2.0f * omega);
 }
 
-void Physics::b2AngularFrequency(float& frequency, float& ratio, float stiffness, float damping, b2Body* bodyA, b2Body* bodyB)
+void Physics::computeAngularStiffness(float &stiffness, float &damping, float frequency, float dampingRatio, b2Body *bodyA, b2Body *bodyB) {
```
slime73, line 442: Braces should be on a newline

src/modules/physics/box2d/Physics.h
```diff
@@ -359,6 +359,10 @@ class Physics : public Module
 	 **/
 	static b2AABB scaleUp(const b2AABB& aabb);
 
+	/**
```
slime73, line 362: This should be filled out (with two `*`'s in the final line to match the other comments).

I don't think each parameter needs to be listed, but a short description would be nice at least

src/modules/physics/box2d/Physics.h
```diff
@@ -368,7 +372,11 @@ class Physics : public Module
 	 * @param bodyA The bodyA of the joint
 	 * @param bodyB The bodyB of the joint
 	 **/
-	static void b2LinearFrequency(float& frequency, float& ratio, float stiffness, float damping, b2Body* bodyA, b2Body* bodyB);
+	static void computeLinearFrequency(float &frequency, float &ratio, float stiffness, float damping, b2Body *bodyA, b2Body *bodyB);
+
+	/**
```
slime73, line 377: Same here with the comment

src/modules/physics/box2d/wrap_Physics.cpp
```diff
@@ -477,12 +477,93 @@ int w_setMeter(lua_State *L)
...
 
 }
+
 int w_getMeter(lua_State *L)
 {
 	lua_pushinteger(L, Physics::getMeter());
 	return 1;
 }
 
+int w_computeLinearStiffness(lua_State *L) {
```
slime73, line 487: Brace style (here and the other functions below)

slime73 (changes requested):

src/modules/physics/box2d/wrap_Physics.cpp
zombrodo: I've become too reliant on auto-formatters :marge:

slime73 (approved):

slime73:
Thanks!

PR #1865, opened 2022-11-22
Make error message in require more descriptive for custom LOVE library loader.

### LINKED ISSUE #1864: Improve `require` Error Message
Currently, it only says "no <module> in LOVE game directories". I think this should give more information about paths it searches too so it says "no file <path/to/module.lua> in LOVE game directories", one for each paths specified from `love.filesystem.getRequirePath` and `love.filesystem.getCRequirePath`.

MikuAuahDark:
This PR adds paths that LOVE tried to load when `require` fails in the error message.

The new output will look like this:
```
Error

main.lua:1: module 'path.to.module' not found:
no field package.preload['path.to.module']
no 'path/to/module' in LOVE game directories:
  tried 'path/to/module.lua'
  tried 'path/to/module/init.lua'
no file 'path/to/module' in LOVE paths:
  tried 'path/to/module.dll' in save directory
no file '.\path\to\module.lua'
no file 'D:\Data\Development\love-dev\install-11\lua\path\to\module.lua'
no file 'D:\Data\Development\love-dev\install-11\lua\path\to\module\init.lua'
no file '.\path\to\module.dll'
no file 'D:\Data\Development\love-dev\install-11\path\to\module.dll'
no file 'D:\Data\Development\love-dev\install-11\loadall.dll'
no file '.\path.dll'
no file 'D:\Data\Development\love-dev\install-11\path.dll'
no file 'D:\Data\Development\love-dev\install-11\loadall.dll'
```

Fixes #1864

### DIFF
--- a/src/modules/filesystem/wrap_Filesystem.cpp
+++ b/src/modules/filesystem/wrap_Filesystem.cpp
@@ -738,12 +738,15 @@ static void replaceAll(std::string &str, const std::string &substr, const std::s
 int loader(lua_State *L)
 {
 	std::string modulename = luax_checkstring(L, 1);
+	std::stringstream errstr;
 
 	for (char &c : modulename)
 	{
 		if (c == '.')
 			c = '/';
 	}
+
+	errstr << "\n\tno '" << modulename << "' in LOVE game directories:";
 
 	auto *inst = instance();
 	for (std::string element : inst->getRequirePath())
@@ -757,11 +760,11 @@ int loader(lua_State *L)
 			lua_pushstring(L, element.c_str());
 			return w_load(L);
 		}
+
+		errstr << "\n\t  tried '" << element << "'";
 	}
 
-	std::string errstr = "\n\tno '%s' in LOVE game directories.";
-
-	lua_pushfstring(L, errstr.c_str(), modulename.c_str());
+	luax_pushstring(L, errstr.str());
 	return 1;
 }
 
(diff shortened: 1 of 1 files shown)


src/modules/filesystem/wrap_Filesystem.cpp
```diff
@@ -828,7 +834,10 @@ int extloader(lua_State *L)
 
 			Filesystem::Info info = {};
 			if (!inst->getInfo(element.c_str(), info) || info.type == Filesystem::FILETYPE_DIRECTORY)
+			{
+				errstr << "\n\t  tried '" << element << "' in save directory";
```
slime73, line 838: I think this isn't guaranteed to only look (or work) in the save directory right now, especially if love 12's new mount APIs have been used.

MikuAuahDark: What would be a good replacement text for those?

slime73: Maybe it could be something like this for the Lua loader:
"No Lua module 'path/to/module' in LOVE filesystem paths:"

And for the C module loader:
"No C module 'path/to/module' in LOVE filesystem paths:"

Does that make sense? I haven't thought about this too hard.

MikuAuahDark: For the C module paths, I was thinking of resolving the fullpath then printing the absolute path, but then I realize `PHYSFS_getRealDir` only works when the file is there (which most likely succeeded when this codepath is reached, defeating the purpose).

PR #1871, opened 2022-12-04
fix: remove usage of setNewFont on error handler

zombrodo:
As it says on the tin.

`love.graphics.setNewFont` is deprecated.

Before:
image

After:
image


MikuAuahDark:
Whoops, forgot to change the error handler when deprecating `love.graphics.setNewFont`. Thank you very much.

PR #1873, opened 2022-12-05
love.sensor module and Joystick sensor functions.

### LINKED ISSUE #1872: Sensor API
As per discussion in Discord, maybe we should use SDL sensor API then deprecate accelerometer-as-joystick option.

MikuAuahDark:
Fixes #1872.

Since I don't have mac to modify the xcode project, the build will fail on those platforms.


src/modules/event/sdl/Event.cpp
```diff
@@ -446,6 +449,35 @@ Message *Event::convert(const SDL_Event &e)
...
+			for (void *s: sensors)
+			{
+				SDL_Sensor *sensor = (SDL_Sensor *) s;
+				SDL_SensorID id = SDL_SensorGetInstanceID(sensor);
+
+				if (e.sensor.which == id)
+				{
+					// Found sensor
+					const char *sensorType = "unknown";
+					sensor::Sensor::getConstant(sensor::sdl::Sensor::convert(SDL_SensorGetType(sensor)), sensorType);
```
slime73, line 467: This should check whether `getConstant` was successful.

src/modules/event/sdl/Event.cpp
```diff
@@ -564,6 +596,26 @@ Message *Event::convertJoystickEvent(const SDL_Event &e) const
...
+#if SDL_VERSION_ATLEAST(2, 0, 14) && defined(LOVE_ENABLE_SENSOR)
+	case SDL_CONTROLLERSENSORUPDATE:
+		stick = joymodule->getJoystickFromID(e.csensor.which);
+		if (stick)
+		{
+			using Sensor = love::sensor::Sensor;
+
+			const char *sensorName = "unknown";
+			Sensor::SensorType sensorType = love::sensor::sdl::Sensor::convert((SDL_SensorType) e.csensor.sensor);
+			Sensor::getConstant(sensorType, sensorName);
```
slime73, line 608: This should check whether getConstant was successful.

slime73 (changes requested):

src/modules/love/callbacks.lua
```diff
@@ -134,6 +137,9 @@ function love.createhandlers()
 				love.audio.setPlaybackDevice()
 			end
 		end,
+		sensorupdate = function (sensorType, x, y, z)
```
slime73, line 140: This (and `joysticksensorupdate` and the event strings) should be past-tense `updated` instead of `update` IMO, to match the other input event names.

PR #1883, opened 2022-12-26
Prefer snprintf over sprintf in OpenGL

devenvexe:
Silences warnings about deprecated sprintf under Xcode 14

### DIFF
--- a/src/modules/graphics/opengl/OpenGL.cpp
+++ b/src/modules/graphics/opengl/OpenGL.cpp
@@ -1937,8 +1937,7 @@ const char *OpenGL::errorString(GLenum errorcode)
 
 	static char text[64] = {};
 
-	memset(text, 0, sizeof(text));
-	sprintf(text, "0x%x", errorcode);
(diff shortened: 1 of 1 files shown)


slime73 (changes requested):
love 11 uses VS2013 on Windows for official builds, which doesn't support snprintf. love 12 uses VS2019 (or later) which does, so I think this change will be good if it's based on the `12.0-development` branch instead of `main`.

devenvexe:
I'll move the PR. Thanks.

MikuAuahDark:
To workaround MSVC 2013, I use something like this 75a3e5db015c05699d0e9f2314801ace505fbd48

slime73:
I think I'd rather have the clean snprintf change in 12.0 rather than some #ifdefs.

devenvexe:
That's cool - and my previous #ifdef won't work of course as I misremembered what is visible when during preprocessor and compilation phases.

devenvexe:
Closing in favor of opening against 12

PR #1884, opened 2022-12-27
Prefer snprintf over sprintf in OpenGL

devenvexe:
Silence warnings against sprintf in OpenGL module. Acceptable since compilers used for love 12 have snprintf.


slime73:
Thanks!

PR #1885, opened 2022-12-27
Vulkan fixes 

nikeinikei:
This pull request fixes a bunch of things that have been found over the last couple of months.

- The apiVersion when creating the vulkan instance should be the highest supported vulkan version. Since VMA (Vulkan Memory Allocator) can already make use of vulkan 1.3 features, it can be hardcoded to VULKAN_API_VERSION_1_3.
- Code styling (spaces -> tabs)
- Some vulkan functions never checked the result value
- love.window.setVsync now has a basic and simple implementation. The updated vsync value gets used in the next frame. A better approach might be to use a newly proposed extension (VK_EXT_swapchain_maintenance1), where the updated vsync value can already get used in the current frame. This needs reevaluation in the future.
- Changing the viewport might now also work on systems that do not throw VK_ERROR_OUT_OF_DATE_KHR (untested)
- Uniform textures and buffers had a bug where the wrong value gets used in the descriptor set. This is now fixed.
- There was a bug where compute shaders would crash if they did not use any local uniform variables (such as float, vec4, mat4, ...). This is now fixed.

There still exists an issue where vsync values > 1 do not work yet. After a bit of research, I did not find a simple solution like there is for OpenGL. One possible solution might be to copy the contents of the framebuffer right before presenting, and then reuse them for the following (vsync - 1) frames. However this seemed out of scope for this pr, since it requires a lot of new code.

One more thing that's on my radar but might need a bigger rewrite and is thus out of the scope of this pull request is how the render passes get managed.
One drawback of the current approach is that the most performant VK_ATTACHMENT_LOAD_OP_CLEAR never gets used.
Also, in some common cases, such as first rendering to a canvas and then rendering the canvas to the frame buffer leads to unnecessary render passes.

A simple example:

```lua
local canvas = love.graphics.newCanvas(...)

function love.draw()
    love.graphics.setCanvas(canvas)
    love.graphics.clear(...)
    love.graphics.draw(...)
    love.graphics.setCanvas()

    love.graphics.draw(canvas)
end
```

Leads to:

1. Render pass for the frame buffer, with only the single command vkCmdClearAttachments
2. Render pass for canvas (with loadOp=load), with the first command being vkCmdClearAttachments
3. Render pass for frame buffer (with loadOp=load)

It is evident that this can be simplified to the equivalent and more performant version:

1. Render pass for canvas (loadOp=clear)
2. Render pass for frame buffer (loadOp=clear)

This is especially useful on mobile devices, where starting render passes are quite expensive (https://developer.android.com/ndk/guides/graphics/design-notes#minimize)


slime73:
> One more thing that's on my radar but might need a bigger rewrite and is thus out of the scope of this pull request is how the render passes get managed.

I was poking at this with some local changes on my old HDD before that died, along with other higher level changes to state changing graphics API implementations in general. I didn't get back to it yet though.

slime73:
> vsync values > 1 do not work yet. After a bit of research, I did not find a simple solution like there is for OpenGL.

I think the Metal backend doesn't support that either yet. It's not a high priority IMO, but if there is a solution that doesn't involve extra textures etc. that'd be nice.

nikeinikei:
> I was poking at this with some local changes on my old HDD before that died, along with other higher level changes to state changing graphics API implementations in general. I didn't get back to it yet though.

Anything specific you had in mind for the higher level changes to the graphics API? I might take a stab at it

PR #1886, opened 2022-12-28
Move all remaining instances of sprintf to snprintf.

devenvexe:
Remove all remaining calls to sprintf in favor of modern snprintf.

### DIFF
--- a/src/libraries/luasocket/libluasocket/auxiliar.c
+++ b/src/libraries/luasocket/libluasocket/auxiliar.c
@@ -51,7 +51,7 @@ int auxiliar_tostring(lua_State *L) {
     lua_pushstring(L, "class");
     lua_gettable(L, -2);
     if (!lua_isstring(L, -1)) goto error;
-    sprintf(buf, "%p", lua_touserdata(L, 1));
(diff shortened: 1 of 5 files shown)


MikuAuahDark:
I'm not too positive of replacing every single sprintf to snprintf on 3rd-party libraries.

slime73:
Yeah, I don't think we want to directly modify third party libraries like this. You could open issues in those libraries' respective upstream source code repositories (assuming they don't have newer versions which address this) - but it's possible they want to support older compilers too.

devenvexe:
Ok, to confirm: third party code isn't modified in the course of Löve development? Alternatively: how do devs approach third party vs. first party code in this project?

slime73:
> third party code isn't modified in the course of Löve development?

That's generally correct, except when there's something preventing the library from working with love. Sometimes we maintain a fork of a library if we have a change that won't be upstreamed in the near future (for example some [PhysFS changes](https://github.com/icculus/physfs/compare/main...slime73:physfs:mountRW) were necessary for a couple 12.0 features. I made an issue report in the PhysFS repository to go along with those changes.)

devenvexe:
I got it. I'll close this PR. Thanks for the clarification.

PR #1887, opened 2022-12-28
Compile with hidden symbol visibility by default.

MikuAuahDark:
This also benefits main branch so I'm targetting main branch instead of 12.0.

### DIFF
--- a/platform/unix/configure.ac
+++ b/platform/unix/configure.ac
@@ -29,6 +29,10 @@ AC_DEFUN([LOVE_MSG_ERROR],
 # C++11 support in cpp11.m4
 ACLOVE_CPP11_TEST
 
+# Add -fvisibility=hidden and -fvisibility-inlines-hidden
+CFLAGS="-fvisibility=hidden $CFLAGS"
+CPPFLAGS="-fvisibility=hidden -fvisibility-inlines-hidden $CPPFLAGS"
+
 # Allow people on OSX to use autotools, they need their platform files
 AC_ARG_ENABLE([osx],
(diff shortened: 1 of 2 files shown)


platform/unix/configure.ac
```diff
@@ -29,6 +29,10 @@ AC_DEFUN([LOVE_MSG_ERROR],
 # C++11 support in cpp11.m4
 ACLOVE_CPP11_TEST
 
+# Add -fvisibility=hidden and -fvisibility-inlines-hidden
+CFLAGS="-fvisibility=hidden $CFLAGS"
+CPPFLAGS="-fvisibility=hidden -fvisibility-inlines-hidden $CPPFLAGS"
```
slime73, line 34: does `fvisibility-inlines-hidden` do anything when `-fvisibility=hidden` is already set?

MikuAuahDark: Yes. From https://gcc.gnu.org/wiki/Visibility

> This causes all inlined class member functions to have hidden visibility, causing significant export symbol table size & binary size reductions though not as much as using `-fvisibility=hidden`. However, `-fvisibility-inlines-hidden` can be used with no source alterations, unless you need to override it for inlines where address identity is important either for the function itself or any function local static data.

MikuAuahDark: Although from what I read it seems to _do nothing_ if `-fvisibility=hidden` is already set, my testing says otherwise as it reduces the size further _slightly_.

slime73: Thanks!

PR #1888, opened 2022-12-31
Allow Canvas:renderTo to accept varargs

steveRoll-git:
In many cases, `Canvas:renderTo` would have been much more useful to me if it could accept additional parameters - without them, you're forced to make a new function if you want it to do something slightly different.
Accepting varargs can let you make the same function do different things without having to create a new one, which can save on a lot of garbage creation.


MikuAuahDark (approved):

slime73:
In the past I've been hesitant to do this for functions that take callbacks because it makes the API less flexible to change, but I think it's fine here.

PR #1892, opened 2023-01-08
Provide PDB with Source Indexing as Artifacts

### LINKED ISSUE #1891: Provide PDB with Source Indexing
Providing PDB in Windows means users with Debugging Tools for Windows can debug problems with LOVE in case of crashes. Having source indexing means that user can pull the source code of LOVE used for compilation of particular (exact) releases.

The PDB can either by hosted at https://love2d.org/pdb (need website access) or distributed as separate archive. However for now I think providing PDB as separate archive is the only way.

MikuAuahDark:
Closes #1891.

To test out if it works:
1. Download the PDBs and put it beside the LOVE executables.
2. Run LOVE (love.exe or lovec.exe) through WinDBG/Visual Studio debugger that run a script which 100% crashes LOVE (i.e. calling `luaopen_love` with NULL `L` through FFI).

I use this script as demonstration.
```lua
local ffi = require("ffi")
local liblove = ffi.load("liblove.dll")
ffi.cdef[[
int luaopen_love(void*)
]]

liblove.luaopen_love(nil)
```

Note that this effectively change the executable name of the LOVE DLL from `love.dll` to `liblove.dll`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -89,18 +89,47 @@ jobs:
         echo nofiles=ignore>> "%GITHUB_OUTPUT%"
         echo moredef=-DLOVE_EXTRA_DLLS=%CD%\angle\libEGL.dll;%CD%\angle\libGLESv2.dll>> "%GITHUB_OUTPUT%"
         exit /b 0
+    - name: Download Windows SDK Setup 10.0.20348
+      run: curl -Lo winsdksetup.exe https://go.microsoft.com/fwlink/?linkid=2164145
+    - name: Install Debugging Tools for Windows
+      id: windbg
+      run: |
+        setlocal enabledelayedexpansion
+        start /WAIT %CD%\winsdksetup.exe /features OptionId.WindowsDesktopDebuggers /q /log %CD%\log.txt
+        echo ERRORLEVEL=!ERRORLEVEL! >> %GITHUB_OUTPUT%
+    - name: Print Debugging Tools Install Log
+      if: always()
+      run: |
+        type log.txt
+        exit /b ${{ steps.windbg.outputs.ERRORLEVEL }}
+    - name: Setup Python 3.10
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.10"
+    - name: Download source_index.py
+      run: curl -Lo source_index.py https://gist.github.com/MikuAuahDark/d9c099f5714e09a765496471c2827a55/raw/df34956052035f3473c5f01861dfb53930d06843/source_index.py
     - name: Clone Megasource
       uses: actions/checkout@v3
       with:
         path: megasource
         repository: love2d/megasource
         ref: 12.x
+    - id: megasource
+      name: Get Megasource Commit SHA
+      shell: python
+      run: |
+        import os
+        import subprocess
+        
+        result = subprocess.run("git -C megasource rev-parse HEAD".split(), check=True, capture_output=True, encoding="UTF-8")
+        commit = result.stdout.split()[0]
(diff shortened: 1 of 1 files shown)


slime73:
This change seems to add several MB to the DLLs of the non-dbg builds.

.github/workflows/main.yml
```diff
@@ -89,18 +89,47 @@ jobs:
         echo nofiles=ignore>> "%GITHUB_OUTPUT%"
         echo moredef=-DLOVE_EXTRA_DLLS=%CD%\angle\libEGL.dll;%CD%\angle\libGLESv2.dll>> "%GITHUB_OUTPUT%"
         exit /b 0
+    - name: Download Windows SDK Setup 10.0.20348
```
slime73, line 92: There are a lot of extra steps here, I'm not confident we could guarantee support for this in the future.

(That's not necessarily a reason to avoid doing it, but it might not last forever if we do it.)

slime73:
> Note that this effectively change the executable name of the LOVE DLL from `love.dll` to `liblove.dll`.

This isn't a dealbreaker but I think it means people who load the love library rather than running the love executable will probably need more platform-specific code to deal with that.

.github/workflows/main.yml
MikuAuahDark: While it's true, unfortunately debugging tools for Windows is not installed in GitHub Actions. One way to avoid it entirely is to write `srctool -r` and `pdbstr` tool equivalent in other languages.

MikuAuahDark:
Regarding the size change, I need to investigate that next week. For people who want to load LOVE as shared library, yeah it can be a dealbreaker. Unfortunately the name change is necessary, otherwise it clash with `love.exe` PDB and `love.dll` PDB and I guess changing `love.exe` is way more dealbreaker than changing `liblove.dll`. This means user who uses LOVE through `require` need to use `require("liblove-love")` instead of `require("love")`.

MikuAuahDark:
After quick searching, looks like CMake `RelWithDebInfo` is not entirely equivalent to `Release` + debug symbols. https://gitlab.kitware.com/cmake/cmake/-/issues/20812

Other than the size increase, it also reduces the performance. So I have to put this on hold for now.

MikuAuahDark:
I redid the PR by forcing `/Zi` C/C++ flags and `/DEBUG` linker flags then compile as Release instead of RelWithDebInfo. The size increase is still there, but it's not as big as RelWithDebugInfo anymore.

love.dll: 6.4 MB -> 8.6 MB
SDL2.dll: 1.5 MB -> 2.1 MB
OpenAL32.dll: 0.9 MB -> 1 MB

Also it no longer renames `love.dll` to `liblove.dll`. WinDBG correctly look for `liblove.pdb` for `love.dll` and source-level debugging still works as before.

slime73:
I don't think I mind a separate downloadable build with that size increase, but IMO the feature is too niche to justify a size increase for all builds.

MikuAuahDark:
I still wonder why the size increase is somewhat proportional to the original binary size though. And weirdly, LuaJIT DLL (lua51.dll) does not have any size increase at all.

I need to investigate it further.

MikuAuahDark:
Updated the LDFLAGS again and there are no more size increase.

Apparently, specifying `/DEBUG` without `/OPT:REF` and `/OPT:ICF` implies `/OPT:NOREF` and `/OPT:NOICF` which increases the binary sizes. Explicitly specifying `/OPT:REF` and `/OPT:ICF` with `/DEBUG` will enable the optimization while also giving PDB.

slime73:
This looks good to me now. Thanks!

PR #1895, opened 2023-01-20
Add 'background' boolean (false by default) to love.window.setMode's …

ITikhonov:
…settings table and love.conf's t.window table.

When background=true, love will no longer automatically raise the window and make it active on creation, but show it in background instead.

-----

I have a specific workflow where i have a love2d window in background while i edit source code in a terminal window and love2d will get automatically brought down and started again once changes are saved. This commit will make it so program's window won't jump into foreground and steal focus from my editor of choice. Not sure if it is useful enough in general but just in case.

Only tested on OSX, but should work elsewhere, probably.

### DIFF
--- a/src/modules/love/boot.lua
+++ b/src/modules/love/boot.lua
@@ -148,6 +148,7 @@ function love.init()
 			centered = true,
 			highdpi = false,
 			usedpiscale = true,
+			background = false,
 		},
 		modules = {
 			data = true,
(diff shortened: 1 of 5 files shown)


MikuAuahDark:
The other-OS equivalent to this is passing `SDL_WINDOW_MINIMIZED` flag to `SDL_CreateWindow`.

zorggn:
> The other-OS equivalent to this is passing `SDL_WINDOW_MINIMIZED` flag to `SDL_CreateWindow` I think.

I don't believe it is equivalent to what the OP wants though; Minimizing would have the window not be shown at all, and from what i understand, they do want the window to be visible, but don't want it to be the active window.

Issue #1904, opened 2023-02-22
Support `vibrate(seconds)` for iOS


ParticleG:
On `iOS 10.0+`, `iPadOS 10.0+` and `Mac Catalyst 13.1+`, it is possible for iPhones and iPads to generate different type of viberations, Not just a fixed 400ms one. Just Mapping seconds to `UIFeedbackGenerator`'s subclasses like this:
```objective-c
if (@available(iOS 10.0, *)) {
	// iOS 10.0 and above
	UIImpactFeedbackGenerator *impact = nil;
	if (seconds >= 0.5 && seconds < 1.5) {
		impact = [[UIImpactFeedbackGenerator alloc] initWithStyle:UIImpactFeedbackStyleLight];
	} else if (seconds >= 1.5 && seconds < 2.5) {
		impact = [[UIImpactFeedbackGenerator alloc] initWithStyle:UIImpactFeedbackStyleMedium];
	} else if (seconds >= 2.5) {
		impact = [[UIImpactFeedbackGenerator alloc] initWithStyle:UIImpactFeedbackStyleHeavy];
	} else {
		return;
	}
	[impact impactOccurred];
	return;
}
```

ParticleG:
And iOS13+ seems have even more options, I'll test them later

ParticleG:
Add the second param to vibrate method with default value `default`, while on iOS `default` means the old fixed 500ms vibration, so it won't affect existing codes:
```c++
// src/modules/system/wrap_System.cpp
int w_vibrate(lua_State *L)
{
	double seconds = luaL_optnumber(L, 1, 0.5);
	const char *mode = luaL_optstring(L, 2, "default");
	instance()->vibrate(seconds, mode);
	return 0;
}
```
And ignore mode param on Android(For now):
```c++
// src/modules/system/System.cpp
void System::vibrate(double seconds, const std::string &mode) const
{
#ifdef LOVE_ANDROID
	love::android::vibrate(seconds);
	LOVE_UNUSED(mode);
#elif defined(LOVE_IOS)
	love::ios::vibrate(seconds, mode);
#else
	LOVE_UNUSED(seconds);
	LOVE_UNUSED(mode);
#endif
}
```
Then process them on iOS:
```objective-c
void vibrate(const double intensity, const std::string &mode)
{
	@autoreleasepool
	{
		if (mode != "default" && @available(iOS 10.0, *)) {
			// iOS 10.0 and above
			UIImpactFeedbackGenerator *impact = nil;
[...]

PR #1905, opened 2023-02-22
Add viberation variations to iOS

ParticleG:
This pull request is for #1904

### DIFF
--- a/src/modules/system/System.h
+++ b/src/modules/system/System.h
@@ -102,9 +102,10 @@ public:
 	/**
 	 * Vibrates for the specified amount of seconds.
 	 *
-	 * @param number of seconds to vibrate.
+	 * @param seconds number of seconds to vibrate.
+	 * @param mode vibration mode.
 	 */
-	virtual void vibrate(double seconds) const;
+	virtual void vibrate(double seconds, const std::string &mode) const;
 
 	/**
 	 * Gets if the user is playing music on background.
--- a/src/common/ios.h
+++ b/src/common/ios.h
@@ -64,9 +64,17 @@ bool openURL(const std::string &url);
 std::string getExecutablePath();
 
 /**
- * Causes devices with vibration support to vibrate for about 0.5 seconds.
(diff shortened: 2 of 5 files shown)


slime73:
Android does expose some haptic feedback preset types, but I'm not sure they really have the exact same meanings as iOS, in which case I suppose iOS and Android are sort of incompatible with each other in terms of the types of haptic feedback they expose...

I'm not sure how I feel about having a function which does radically different things (and effectively accepts different parameters) depending on the platform. I guess it was like that a little bit already, but this makes the difference even bigger.

src/modules/system/System.h
```diff
@@ -102,9 +102,10 @@ class System : public Module
 	/**
 	 * Vibrates for the specified amount of seconds.
 	 *
-	 * @param number of seconds to vibrate.
+	 * @param seconds number of seconds to vibrate.
+	 * @param mode vibration mode.
 	 */
-	virtual void vibrate(double seconds) const;
+	virtual void vibrate(double seconds, const std::string &mode) const;
```
slime73, line 108: I think this should use an enum here (like the PowerState enum in this file, if you need an example to work from) instead of passing the string all the way down to the platform implementation.

slime73 (changes requested):

ParticleG:
> Android does expose some haptic feedback preset types, but I'm not sure they really have the exact same meanings as iOS, in which case I suppose iOS and Android are sort of incompatible with each other in terms of the types of haptic feedback they expose...
> 
> I'm not sure how I feel about having a function which does radically different things (and effectively accepts different parameters) depending on the platform. I guess it was like that a little bit already, but this makes the difference even bigger.

Well, currently Android would switch between haptic vibration and normal long vibration depends on how long the vibration is, while iOS could only provide a constant 500ms vibration, and that's a big difference. Different platforms accept different params do sounds bad, so I'm considering tweaking iOS's vibration only with the `seconds` param, as I mentioned in the #1904 . Maybe test it on 4 ~ 5 iphone/android phone I think.

PR #1906, opened 2023-02-22
graphics: fix for anti-parallel line segments

### LINKED ISSUE #1894: Edge case in both miter/bevel linejoin modes
This is such a minor bug that it feel pointless to report, but it could confuse/upset someone new to love2d/coding and that would be unfortunate.

For fun I'm working on a love2d controller of the derail valley dispatcher mod. While working on rendering the tracks I ended up with a set of points for a track segment that when rendered cause what should be a short section of track to be a line that runs off screen in both directions. When you set the linejoin mode to bevel one side corrects but the other does not. If you set linejoin mode to none then it renders fine. You can also reduce the decimal precision down to remove the errors but the points listed are generated dynamically while zooming/panning the map.

I have extracted the points and they still reliably glitched for me on both my Linux and Windows machines. I should add that this error occurs when the map is fully zoomed out, that is why this line segment is so short. This is on version 11.3.

function love.draw()
	love.graphics.setLineJoin( "miter" )
	--love.graphics.setLineJoin( "bevel" )
	--love.graphics.setLineJoin( "none" )
	love.graphics.line({ 121.852846370545,109.87029686964,
		121.672390327231,109.35841847852,
		121.491941615924,108.84652359026,
		121.308234826955,108.32542462031,
		121.311489571886,108.33464519913 })
end

Screenshots of the errors with the different linejoin modes
errors

Zipped version of the code above
[main.zip](https://github.com/love2d/love/files/10407643/main.zip)

scurest:
Fixes #1894.

Correctly handles the case where a line doubles back on itself. For both miter and bevel, the line will be drawn like this (angle increased for readability; in reality the angle pqr would be much smaller of course)

Because the normal flips here, the triangle strip would twist. So to draw this we "contain" the twist in an invisible zero-size quad

join2

Here's an example showing the importance of the twist correction

```lua
function love.draw()
    love.graphics.setLineWidth(20)
    love.graphics.setLineJoin("miter")
    love.graphics.line({
        190, 200,
        200, 200,
        100, 200
    })
    love.graphics.setLineJoin("bevel")
    love.graphics.line({
        100, 300,
        200, 300,
        190, 300
    })
end
```

### DIFF
--- a/src/modules/graphics/Polyline.cpp
+++ b/src/modules/graphics/Polyline.cpp
@@ -157,7 +157,7 @@ void NoneJoinPolyline::renderEdge(std::vector<Vector2> &anchors, std::vector<Vec
  *    (q + w/2 * ns) + lambda * (q - p) = (q + w/2 * nt) + mu * (r - q)   (u1)
  *    (q - w/2 * ns) + lambda * (q - p) = (q - w/2 * nt) + mu * (r - q)   (u2)
  *
- * with nt,nt being the normals on the segments s = p-q and t = q-r,
+ * with ns,nt being the normals on the segments s = p-q and t = q-r,
  *
  *    ns = perp(s) / |s|
  *    nt = perp(t) / |t|.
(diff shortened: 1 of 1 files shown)


slime73:
I tested a bit with a few different uses of lines and this seems to hold up well. Thanks!

PR #1926, opened 2023-05-09
Added variants for peer:send and host:broadcast which accepts light userdata and size

EngineerSmith:
Was #1791 but rewritten to exclude the include of `love::data` by using light userdata and a size

This adds the following variants for `peer:send` and `host:broadcast` while preserving the previous function arguments
`peer:send(light userdata pointer, number size, [number channel, [string flag]])`
`host:broadcast(light userdata pointer, number size, [number channel, [string flag]])`

Working example
```lua
-- client
local enet = require("enet")
local host = enet.host_create()
local server = host:connect("localhost:10001", 64, 0)
host:service(100)

-- populate byteData
local byteData = love.data.newByteData(5)
local ffi = require("ffi")
local ptr = ffi.cast("void*", byteData:getFFIPointer())
ffi.copy(ptr, "hello")

server:send(byteData:getPointer(), byteData:getSize(), 11, "reliable") -- send data

love.update = function()
  host:service(10)
end
```
```lua
-- server
local enet = require("enet")
local host = enet.host_create("*:10001", 64, 0)

local text, channel
love.update = function()
  local event = host:service(10)
  if event and event.type == "receive" then
    text = event.data
    channel = event.channel
  end
end

love.draw = function()
-- prints 'hello' once client joins and the channel it was sent on
  love.graphics.print(tostring(text).."\nChannel: "..tostring(channel))
end
```

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -79,6 +79,7 @@ Released: N/A
 * Added love.sensor module.
 * Added love.sensorupdated callback.
 * Added love.joysticksensorupdated callback.
+* Added variant for enet peer:send and host:broadcast which accepts a pointer (light userdata) and a size.
 
 * Changed love.filesystem.exists to no longer be deprecated.
 * Changed the default font from Vera size 12 to Noto Sans size 13.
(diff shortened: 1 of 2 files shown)


slime73:
Looks good, thanks!

Issue #1047, opened 2015-05-03
os.execute(command) always return -1


slime73:
**[Original report](https://bitbucket.org/rude/love/issue/1042) by João Lopes (Bitbucket: [linux-man](https://bitbucket.org/linux-man), GitHub: [linux-man](https://github.com/linux-man)).**

----------------------------------------

Love 0.9.2 on Ubuntu 64:
os.execute() return 0.
os.execute(command) return -1 despite if the command is successful or fail.
Is luajit breaking lua compatibility? Or is a love bug?

PR #1928, opened 2023-05-14
Reap zombie processes left behind by openURL

pfirsich:
This was attempted to be fixed in f7d226d but sadly broke os.execute (#1047), so it was later removed in a4a62f3.
This is a very similar fix, but we make sure to only waitpid for processes that we are supposed to reap.

Without this fix an example program like this would leave behind zombie processes:
```lua
function love.load()
    love.system.openURL("https://theshoemaker.de")
    love.system.openURL("https://google.de")
end
```
```bash
$ ps --ppid (pidof love)
    PID TTY          TIME CMD
 133235 pts/2    00:00:00 xdg-open <defunct>
 133236 pts/2    00:00:00 xdg-open <defunct>
```
With the fix, those zombie processes are gone.
I also made sure that this doesn't also break os.execute. This example program:
```lua
function love.load()
    love.system.openURL("https://theshoemaker.de")
    love.system.openURL("https://google.de")
    print(os.execute("false"))
    print(os.execute("true"))
end
```
prints
```
256
0
```

I am not sure if my comment in the code is too verbose, but considering it took me a while to figure out what went on there in the past, I thought it would be better to write too much than too little. Let me know if anything needs changing.

Also I am not sure if that fix is even worth it. It's not very clean to leave them behind and they might be annoying, but the typical löve program does not open enough URLs for this to become a real problem.

### DIFF
--- a/src/modules/system/System.cpp
+++ b/src/modules/system/System.cpp
@@ -53,6 +53,29 @@
 #endif
 #endif
 
+#if defined(LOVE_LINUX)
+#include <unordered_set>
+
+static std::unordered_set<pid_t>& getSpawnedPids()
+{
+	static std::unordered_set<pid_t> spawnedPids;
+	return spawnedPids;
+}
+
+static void sigchldHandler(int signo, siginfo_t* info, void* /*context*/)
+{
+	// We need to be selective when reaping child processes, because simply calling
+	// `waitpid` on every child process that terminated will break any other calls
+	// to `waitpid`.
+	// Particularly `os.execute` (internally: `system`) will always return -1, because
+	// their call to `waitpid` returned `ECHILD`, since we already reaped the process.
+	// This was reported in #1047 and fixed with a similar approach in 30ff7d4 and f7d226d
+	// and then later removed in a4a62f3.
+	if (getSpawnedPids().erase(info->si_pid) > 0)
+		::waitpid(info->si_pid, nullptr, 0);
+}
+#endif
+
 namespace love
 {
 namespace system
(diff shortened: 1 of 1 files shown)


src/modules/system/System.cpp
```diff
@@ -53,13 +53,43 @@
...
+#if defined(LOVE_LINUX)
+#include <unordered_set>
+
+static std::unordered_set<pid_t>& getSpawnedPids()
+{
+	static std::unordered_set<pid_t> spawnedPids;
+	return spawnedPids;
+}
+
+static void sigchldHandler(int signo, siginfo_t* info, void* /*context*/)
```
slime73, line 65: I'm very unfamiliar with these APIs so this might be a silly question: is it possible for this callback to be called on other threads, or is it guaranteed to only be called on the main thread?

If it can be called on other threads there'll need to be some extra thread safety (maybe a mutex) when modifying `spawnedPids`.

src/modules/system/System.cpp
```diff
@@ -53,13 +53,43 @@
...
 {
 
 System::System()
 {
+#if defined(LOVE_LINUX)
+	struct sigaction sa;
+	sa.sa_sigaction = sigchldHandler;
+	sigemptyset(&sa.sa_mask);
+	sa.sa_flags = SA_RESTART | SA_SIGINFO;
+	sigaction(SIGCHLD, &sa, nullptr);
```
slime73, line 91: Does this get cleaned up / reset when the `System` class is destroyed? e.g. what happens when `love.event.quit("restart")` is called (which keeps the process running but starts new instances of love's modules)?

slime73:
love now uses SDL_OpenURL on all platforms, I think it'll fix remaining issues in love.system.openURL on Linux.

PR #1930, opened 2023-05-17
Add PHYSFS_Io StripSuffixIo to skip codesign signatures

pfirsich:
When a fused löve executable is signed with codesign on Windows some data is appended to the executable, preventing physfs from finding the zip file at the end of the file.
To help physfs out a bit, we try to find the zip ourselves in the case of regular mount failure by looking for it's characteristic end of central directory record and, once found, pretend to physfs that the file is shorter than it actually is and ends at the end of the zip file.

------

This is another one of those PRs that don't try to say "please merge", but rather a "do you possibly want this?". 

There were questions on the Discord about issues with loading fused, signed Windows executables. I thought it's possible to do this with a custom PHYSFS_Io, so I did it. It seemed like there were other people possibly considering signing their Windows executables too. 

A version of my changes were tested by KayleMaster with his game Stone Kingdoms. I also verified on Linux by appending a few KB of random data to fused executables. I tested it non-fused too with existing and not existing arguments. 

I think generally I am not sure if the volume and complexity is worth the added value. Also I am not sure if the PhysfsIo wrapper I made might be a bit out of place and I should align more with the `PHYSFS_Io` @MikuAuahDark  has built for Android. One real concern is also that this IO might be less robust or efficient than the built-in ones, but I think to really do it well, you would need separate implementations for each platform, which is probably not worth the effort, especially considering this is only used as a fallback. KayleMaster said that he will compare loading times with/without my changes soon. Maybe this will give some indication as well.

### DIFF
--- a/src/modules/filesystem/physfs/PhysfsIo.h
+++ b/src/modules/filesystem/physfs/PhysfsIo.h
@@ -1,0 +1,160 @@
+/**
+ * Copyright (c) 2006-2023 LOVE Development Team
+ *
+ * This software is provided 'as-is', without any express or implied
+ * warranty.  In no event will the authors be held liable for any damages
+ * arising from the use of this software.
+ *
+ * Permission is granted to anyone to use this software for any purpose,
+ * including commercial applications, and to alter it and redistribute it
+ * freely, subject to the following restrictions:
+ *
+ * 1. The origin of this software must not be misrepresented; you must not
+ *    claim that you wrote the original software. If you use this software
+ *    in a product, an acknowledgment in the product documentation would be
+ *    appreciated but is not required.
+ * 2. Altered source versions must be plainly marked as such, and must not be
+ *    misrepresented as being the original software.
+ * 3. This notice may not be removed or altered from any source distribution.
+ **/
+
+#ifndef LOVE_FILESYSTEM_PHYSFS_PHYSFSIO_H
+#define LOVE_FILESYSTEM_PHYSFS_PHYSFSIO_H
+
+#include <cstdint>
+#include <cstdio>
+#include <string>
+
+#include "libraries/physfs/physfs.h"
+
+namespace love
+{
+namespace filesystem
+{
+namespace physfs
+{
+
+template <typename Derived>
+struct PhysfsIo : PHYSFS_Io {
+protected:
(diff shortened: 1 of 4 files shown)


pfirsich:
It seems there are undefined references in the iOS build, so the new source files probably need to be added to another list, but I don't know which ones. If there is interest in this PR, let me know where I need to add them.

slime73:
> This is another one of those PRs that don't try to say "please merge", but rather a "do you possibly want this?".

I think this approach is pretty reasonable. The `StripSuffixIo` code needs a bit of specialized knowledge to write/maintain but that's sort of unavoidable I guess.

> It seems there are undefined references in the iOS build, so the new source files probably need to be added to another list, but I don't know which ones. If there is interest in this PR, let me know where I need to add them.

Don't worry about that part, I'll handle it after the merge since it sort of needs Xcode in order to generate the right additions to the xcodeproj.

src/modules/filesystem/physfs/PhysfsIo.h
```diff
@@ -0,0 +1,160 @@
...
+		delete derived(io);
+	}
+
+	static Derived* derived(PHYSFS_Io* io)
+	{
+		return static_cast<Derived*>(reinterpret_cast<PhysfsIo*>(io->opaque));
+	}
+};
+
+struct StripSuffixIo : public PhysfsIo<StripSuffixIo> {
```
slime73, line 112: I think the split `PhysfsIo` template / `StripSuffixIo` definitions could be merged together for some simpler code here (and could be split apart again if we ever need to in the future), but I don't really have strong feelings about needing it simplified before merging this PR.

MikuAuahDark: I think the split between `PhysfsIo` template class is useful. I could re-use it for Android AAsset archiver backend.

pfirsich: Yeah, I made it a separate class so it can be used for the PhysfsIOs that are already used on Android, if there is interest, which it seems there is :)

slime73: Alright, that makes sense. Thanks!

PR #1931, opened 2023-05-18
Alternative Fusing Method using Windows Resource

MikuAuahDark:
This PR must be discussed thoroughly before being merged, since this PR is meant to compete with #1930. Criticism and suggestions are welcome.

-----

This PR adds "alternative" fusing method in Windows that's compatible when the executable is signed. Traditional fusing (`copy /b love.exe+game.love game.exe`) does not work when the resulting executable (`game.exe`) is signed, resulting in no-game screen after signing.

To use this alternative fusing method, use a resource editor to `love.exe` or `lovec.exe` and add a new resource with type of `RCDATA` (Binary Resources) called "`GAME`". The resource contents should be an archive supported by PhysFS containing all your game just like an ordinary .love file has.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -324,6 +324,8 @@ set(LOVE_SRC_COMMON
 	#src/common/Vector.cpp # Vector.cpp is empty.
 	src/common/Vector.h
 	src/common/version.h
+	src/common/win32.cpp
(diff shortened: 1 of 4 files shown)


slime73:
I think I'm leaning towards the approach in #1930 since it needs less setup from users.

Issue #1932, opened 2023-05-19
Incorrect and unstable MD5 hashing.


ImagicTheCat:
Hi,

main.lua:
```lua
local data = "test"
local i = 0
local hash
function love.update(dt)
  i = i+1
  local new_hash = love.data.hash("md5", data)
  assert((not hash) or (hash == new_hash))
  hash = new_hash
  print(i, data, love.data.encode("string", "hex", hash))
end
```

Will throw an error for `md5`, but will be fine for other hash functions.

I suspected that there was some corruption or uninitialized memory, so I printed the content of the prepared data.

```diff
diff --git a/src/modules/data/HashFunction.cpp b/src/modules/data/HashFunction.cpp
index 643d180b..bd2e9c47 100644
--- a/src/modules/data/HashFunction.cpp
+++ b/src/modules/data/HashFunction.cpp
@@ -95,6 +95,10 @@ class MD5 : public HashFunction
 		*((uint64*) &padded[paddedLength]) = length * 8;
 		paddedLength += 8;

+    for(int i = 0; i < paddedLength; i++)
+      printf("%d ", (int)padded[i]);
+    printf("%s", "\n");
+
 		for (uint64 i = 0; i < paddedLength; i += 64)
 		{
 			uint32 *chunk = (uint32*) &padded[i];
```

C++ is not part of my methodology and I don't understand clearly what the code is doing, but with that code the hash is correct and stable, thus I suspect that there are some undefined behaviors going on. Maybe a case of strict aliasing violation.

It happens with LÖVE 11.4: the binary provided by Arch Linux `love 11.4-1` and a manually compiled version from tag `11.4`.

ImagicTheCat:
(After a bit of [C homework](https://en.cppreference.com/w/c/language/object))

As I assumed, the hash implementations are violating strict aliasing. It would be better to rely on a good hashing dependency, easy to integrate in LÖVE. But I suppose it wasn't, maybe still isn't, easy to find one.

I made a fix for the MD5 function. It assumes that `paddedLength` is a multiple of 4, which I didn't verified yet.

```diff
diff --git a/src/modules/data/HashFunction.cpp b/src/modules/data/HashFunction.cpp
index 34d9e008..307e1dc9 100644
--- a/src/modules/data/HashFunction.cpp
+++ b/src/modules/data/HashFunction.cpp
@@ -86,18 +86,19 @@ class MD5 : public HashFunction
 		if (paddedLength % 64 > 56)
[...]

PR #1935, opened 2023-05-24
Hash function fixes.

ImagicTheCat:
Following https://github.com/love2d/love/issues/1932.

Here are my changes for MD5 before I proceed to do the other functions to preventively check if this is fine to you.

I have written this small test for the purpose of the contribution:
```lua
local rand = require "openssl.rand"
local digest = require "openssl.digest"

local function check_func(name, data)
	local lhash, rhash = digest.new(name):final(data), love.data.hash(name, data)
	assert(lhash == rhash)
end

function love.update(dt)
	local size = rand.uniform(5*2^20)
	local data = rand.bytes(size)
	print("check data("..size..")...")
	check_func("md5", data)
	check_func("sha1", data)
	check_func("sha224", data)
	check_func("sha256", data)
	check_func("sha384", data)
	check_func("sha512", data)
end
```

### DIFF
--- a/src/modules/data/HashFunction.cpp
+++ b/src/modules/data/HashFunction.cpp
@@ -455,31 +455,26 @@ public:
 		else
 			memcpy(intermediates, initial512, sizeof(intermediates));
 
-		//Do the required padding
-		uint64 paddedLength = length + 1; //Consider the appended bit
-		if (paddedLength % 128 < 112)
-			paddedLength += 112 - paddedLength % 128;
-		if (paddedLength % 128 > 112)
-			paddedLength += 240 - paddedLength % 128;
+		// Compute final padded length, accounting for the appended bit (byte) and size
+		uint64 paddedLength = extend_multiple(length + 1 + 16, 128);
 
-		uint8 *padded = new uint8[paddedLength + 16];
-		paddedLength += 8;
+		uint64 *padded = new uint64[paddedLength / 8];
 		memcpy(padded, input, length);
-		memset(padded + length, 0, paddedLength - length);
-		padded[length] = 0x80;
+		memset(((uint8*)padded) + length, 0, paddedLength - length);
+		*(((uint8*)padded) + length) = 0x80; // append bit
 
-		// Now we need the length in bits (big endian), note we only write a 64-bit int, so
+		// Append length in bits (big endian), note we only write a 64-bit int, so
 		// we have filled the first 8 bytes with zeroes
-		length *= 8;
-		for (int i = 0; i < 8; ++i, ++paddedLength)
-			padded[paddedLength] = (length >> (56 - i * 8)) & 0xFF;
+		uint64 bit_length = length * 8;
+		for (int i = 0; i < 8; ++i)
+			*(((uint8*)padded) + (paddedLength - 8 + i)) = (bit_length >> (56 - i * 8)) & 0xFF;
 
 		// Allocate our extended words
 		uint64 words[80];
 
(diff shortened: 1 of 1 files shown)


ImagicTheCat:
Done. Some redundant code could be simplified.

src/modules/data/HashFunction.cpp
```diff
@@ -79,25 +86,22 @@ class MD5 : public HashFunction
...
 		uint32 d0 = 0x10325476;
 
-		//Do the required padding (MD5, SHA1 and SHA2 use the same padding)
-		uint64 paddedLength = length + 1; //Consider the appended bit
-		if (paddedLength % 64 < 56)
-			paddedLength += 56 - paddedLength % 64;
-		if (paddedLength % 64 > 56)
-			paddedLength += 120 - paddedLength % 64;
+		// Compute final padded length, accounting for the appended bit (byte) and size
+		uint64 paddedLength = extend_multiple(length + 1 + 8, 64);
```
slime73, line 90: I think the `+ 8` should be outside of the `extend_multiple` function here (like in the old code), otherwise part of it could be absorbed by the padding even though all 8 bytes need to be available at the end, if I'm understanding right.

ImagicTheCat: This function extends/increases the value of `length + 1 + 8` so that it becomes a multiple of 64, thus the result cannot be smaller than `length + 1 + 8` , thus  it cannot be absorbed (if `extend_multiple` is correct).

It is effectively increasing the value by the amount of padding zeroes.

To my understanding, the old code was making `length + 1` congruent to 56 modulo 64, implying `(length + 1) % 64 == 56`, to then add the 8 bytes, making it a multiple of 64.

To clarify, I directly computed the `paddingLength` to be a multiple of 64, rather than thinking about those intermediate computations.

slime73: Admittedly I don't fully understand the original code, but my assumption would be that the padding exists in the first place so the code can store data there. So if the extra 8 bytes is part of the same input as the original length to produce the padded length, then the algorithm's code might use the extra 8 bytes' memory when it shouldn't (depending on what the original length was), right?

ImagicTheCat: Unless I'm missing something, yes this is intended. The final hashed payload is `message + 0x80 + 0x00... + size(8 bytes)` with a length multiple of 64. The old code was adding `8` to `paddedLength` afterwards. It's one of the reasons I removed it, the `paddedLength` was changing progressively, hard to analyze.

Maybe `paddedLength` is a confusing name, but I don't really like `finalLength`.

ImagicTheCat: Maybe `payloadLength` would be better ?

src/modules/data/HashFunction.cpp
```diff
@@ -79,25 +86,22 @@ class MD5 : public HashFunction
...
-		//Now we need the length in bits
-		*((uint64*) &padded[paddedLength]) = length * 8;
-		paddedLength += 8;
+		// Append length in bits
+		uint64 bit_length = length * 8;
+		memcpy(((uint8*)padded) + paddedLength - 8, &bit_length, 8);
 
-		for (uint64 i = 0; i < paddedLength; i += 64)
+		// Process chunks
+		for (uint64 i = 0; i < paddedLength/4; i += 16)
```
slime73, line 102: Won't this loop iterate over different values than before, since paddedLength means a different thing now?

PR #1944, opened 2023-06-15
Fix #1943 - Duplicate frames being skipped when decoding Theora

### LINKED ISSUE #1943: Identical frames in videos are skipped
Hey all,

When playing a video using love.graphics.newVideo(), if the video contains identical consecutive frames, only the first frame is shown and the rest are skipped (I would guess that the encoder only keeps the first identical frame an indicates how long it should be on screen for). This is most noticeable when audio is included, as the audio is not skipped and becomes desynced.

I have created an example project to demonstrate this issue. The video contains three titles, the first is 5 seconds long with no animation, the second is 1 second long with no animation, and the final one is 5 seconds long and includes dots to make every frame different so that it plays back normally.

[love-video-bug-repro.zip](https://github.com/love2d/love/files/11740436/love-video-bug-repro.zip)

test.ogv is encoded using ffmpeg with the following command:
`ffmpeg -r 24 -i encode-from.mp4 -r 24 -c:v libtheora -q:v 10 -c:a libvorbis output.ogv -y`

Thanks for all the good work that you do!

MellowArpeggiation:
Uses the granule position encoded on each packet to update the granule position of the decoder using `th_decode_ctl()`. A bit of reordering of the remaining code ensures that the video frames come out at the exact correct time.

Works with both the test case provided and various other videos. Used Firefox as a reference player due to both VLC and Chromium also exhibiting playback issues!

### DIFF
--- a/src/modules/video/theora/TheoraVideoStream.cpp
+++ b/src/modules/video/theora/TheoraVideoStream.cpp
@@ -229,17 +229,24 @@ void TheoraVideoStream::threadedFillBackBuffer(double dt)
 			failedSeek = true;
 		}
 
+		if (packet.granulepos > 0) {
+			th_decode_ctl(decoder, TH_DECCTL_SET_GRANPOS, &packet.granulepos, sizeof(packet.granulepos));
+		}
+
+		// TODO: Should be checking the result of th_decode_packetin
 		th_decode_ycbcr_out(decoder, bufferinfo);
 		hasFrame = true;
 
-		ogg_int64_t granulePosition;
-		do
-		{
-			if (demuxer.readPacket(packet))
-				return;
-		} while (th_decode_packetin(decoder, &packet, &granulePosition) != 0);
+		// Decode the current packet for the frame
+		ogg_int64_t decoderPosition;
+		th_decode_packetin(decoder, &packet, &decoderPosition);
+
+		// Prepare the next packet for the next frame
+		if (demuxer.readPacket(packet))
+			return;
+
 		lastFrame = nextFrame;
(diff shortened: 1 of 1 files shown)


slime73:
It looks like rewinding (and maybe other sorts of seeking) is more broken now than before, it seems to freeze the video when I try.

MellowArpeggiation:
When working in 11.4 I did notice that `seek()` was only reliable forwards so I always called `rewind()` first before calling seek (and did the same with the few tests I did in this branch) and elected not to focus on that for this fix, but since I've already learned the ins and outs of this area I'll make some test cases for seeking backwards and forwards and see what I can do!

slime73:
While I was testing this patch, I called `rewind()` while the video was still playing and it broke video playback (but audio playback worked as expected), whereas both video and audio rewound in that situation in 11.4's code.

MellowArpeggiation:
I'm not able to repro broken video playback with this:

```lua
function love.load()
    video = love.graphics.newVideo("test.ogv")
    love.window.setMode(video:getWidth(), video:getHeight())
    video:play()
end

function love.update(dt)
    if not hasRewind and love.timer.getTime() > 5 then
        video:pause()
        video:rewind()
        video:play()
        hasRewind = true
    end
end

function love.draw()
    love.graphics.draw(video)
end
```

Can you send the video you're testing with?

MellowArpeggiation:
Nevermind, managed to repro, it doesn't occur in the Debug build, only in Release

slime73:
The latest commit fixes that issue for me, and I haven't found any other regressions so far. Thanks for investigating and fixing these problems!

PR #1945, opened 2023-06-16
Enable linting and configure clang-format

mmatheson:
I agree with the motivation behind the coding style for this repo [here](https://love2d.org/wiki/Code_Style), and realized most (if not all) of the recommendations there can be automatically enforced and fixed.

I added a .clang-tidy config that matches the coding style referenced in the docs, and used Trunk to enable and run all the relevant linters.

I'm not trying to change the coding style in this repo, I don't particularly care one way or another - but a consistent, enforced coding style is very valuable. Happy to iterate on this PR with the maintainer to get a consensus on the .clang-format config and the other linters that were auto-enabled.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,22 @@
+BasedOnStyle: Google
+ColumnLimit: 100
+---
+Language: Cpp
+DerivePointerAlignment: false
+BreakBeforeBraces: Allman
+PackConstructorInitializers: Never
+NamespaceIndentation: None
+BreakInheritanceList: BeforeColon
+SpaceBeforeParens: Custom
+SpaceBeforeParensOptions:
+  AfterFunctionDeclarationName: false
+  AfterFunctionDefinitionName: false
+  AfterControlStatements: true
+IndentCaseBlocks: false
+PointerAlignment: Right
+UseTab: ForIndentation
+SpaceAfterCStyleCast: true
+SpacesInCStyleCastParentheses: false
+AllowShortIfStatementsOnASingleLine: false
+---
+Language: Proto
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,4 @@
 BasedOnStyle: Google
-ColumnLimit: 100
 ---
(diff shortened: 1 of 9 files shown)


.clang-format
```diff
@@ -0,0 +1,22 @@
+BasedOnStyle: Google
+ColumnLimit: 100
```
slime73, line 2: love doesn't have a hard character limit in its code.

.clang-format
slime73, line 1: This needs to only apply to `src/common` and `src/modules`.

slime73: I tried using this branch with Visual Studio and it still tries to format files in `src/libraries`. Also VS2019 gets stuck trying to format any file I try.

svgo.config.js
slime73, line 1: What does this do?

.shellcheckrc
slime73, line 1: What does this do?

.markdownlint.yaml
slime73, line 1: This looks like a documentation / .md style enforcer, do we need that? I expected this PR to just be about code style.

.clang-format
mmatheson: Are you developing on Windows? What commands were you running?

slime73: Yes (well, I have Windows and macOS machines I develop on). I didn't use any commands, Visual Studio automatically recognizes clang-format files and integrates their formatting configuration into its workflow.

In my opinion it's a lot more important to get the editing workflow part right before thinking about whether build-time code style checks make sense for this project, since editors are where people actually make code changes in the first place.

slime73:
Is this PR still being worked on? As I mentioned in a previous comment I think the best thing for this codebase would be to start with getting formatting choices working with IDEs, and then *maybe* add post-push/PR formatting checks once we've been using that for a while. If that's not going to happen in this PR I'll close it.

slime73:
It looks like this isn't going anywhere so I'll close this, but feel free to make a new PR based on my earlier comments here.

PR #1953, opened 2023-07-10
fix reference to macos.mm

hgoschuetz:


CMakeLists.txt
```diff
@@ -328,7 +328,7 @@ set(LOVE_SRC_COMMON
 
 if (APPLE)
 	set(LOVE_SRC_COMMON ${LOVE_SRC_COMMON}
-		src/common/macosx.mm
+		src/common/macos.mm
```
slime73, line 331: This file is still `macosx.mm` in the main branch, it's only renamed to `macos.mm` in `12.0-development`. Did you mean to make the PR based against that branch instead of `main`?

hgoschuetz: Yes, I did. Sorry for the inconvenience.

PR #1964, opened 2023-09-01
Add SDF HintingMode

Labrium:
Freetype 2.11.1 allows this new scalable font render mode, so this PR adds the ability to use it in Löve.


slime73:
Can you make this based on the `12.0-development` branch of love instead of `main`? Unfortunately you might have to recreate the pull request once you do that.

Labrium:
Sure, I realized that my build system had some old files still cached, so I have to make a few more changes anyway.

PR #1974, opened 2023-10-04
basic API test suite

ellraiser:
Initial commit of a basic test framework, see readme.md in /testing for more

Most modules are covered with basic unit tests, and there's an example test for a graphics draw (rectangle) and object (File) - for object tests doing more scenario based so we can check multiple things together, if style is fine will carry on working through the other graphics methods + object types (which is the bulk of the work tbh)


testing/classes/TestMethod.lua
```diff
@@ -0,0 +1,359 @@
...
+    self.testmodule:log(
+      self.testmodule.colors[self.result.result],
+      '  ' .. tested .. matching,
+      ' ==> ' .. self.result.result .. ' - ' .. endtime .. 'ms ' ..
+      self.result.total .. msg
+    )
+  end
+
+
+}
```
slime73, line 359: git likes it if you have a newline at the end of files

testing/classes/TestSuite.lua
```diff
@@ -0,0 +1,159 @@
...
+  end,
+
+
+  -- @method - TestSuite:printResult()
+  -- @desc - prints the result of the whole test suite as well as writes
+  --         the XML + HTML of the testsuite output
+  -- @return {nil}
+  printResult = function(self)
+    local finaltime = tostring(math.floor(self.time*1000))
+    if string.len(finaltime) == 1 then finaltime = '   ' .. finaltime end
```
slime73, line 128: `#finaltime` is generally more idiomatic than `string.len(finaltime)`

testing/tests/audio.lua
```diff
@@ -0,0 +1,296 @@
...
+  })
+  test:assertEquals(1, #love.audio.getActiveEffects(), 'check 1 effect running')
+  test:assertEquals('testeffect', love.audio.getActiveEffects()[1], 'check effect details')
+end
+
+
+-- love.audio.getActiveSourceCount
+love.test.audio.getActiveSourceCount = function(test)
+  -- tests
+  test:assertNotEquals(nil, love.audio.getActiveSourceCount(), 'check not nil')
```
slime73, line 21: I think the `assertEquals` below (and in many other tests) already covers situations where the function returns nil.

testing/tests/audio.lua
```diff
@@ -0,0 +1,296 @@
...
+  test:assertEquals(0, love.audio.getActiveSourceCount(), 'check 0 by default')
+  local testsource = love.audio.newSource('resources/click.ogg', 'static')
+  test:assertEquals(0, love.audio.getActiveSourceCount(), 'check not active')
+  love.audio.play(testsource)
+  test:assertEquals(1, love.audio.getActiveSourceCount(), 'check now active')
+  love.audio.pause()
+  testsource:release()
```
slime73, line 28: It'd be nice to avoid having to use `object:release()` in every test that uses an object. If you `collectgarbage("collect")` between every test does it significantly increase the time it takes for all tests to run? If not I think I'd prefer something like that instead of having release() inside tests.

testing/tests/window.lua
```diff
@@ -0,0 +1,336 @@
...
+love.test.window.getDPIScale = function(test)
+  test:assertNotEquals(nil, test, 'check not nil')
+end
+
+
+-- love.window.getDesktopDimensions
+-- @NOTE dependent on hardware so best can do is not nil
+love.test.window.getDesktopDimensions = function(test)
+  local w, h = love.window.getDesktopDimensions()
+  test:assertNotEquals(nil, w, 'check not nil')
```
slime73, line 35: It might be nice to have a helper function like `assertNotNil` so we don't have to write or copy/paste the 'check not nil' string everywhere

slime73:
I think the place I'd expect to have object tests is inside the file for the module that owns the object, e.g. `Transform` and `RandomGenerator` tests would be in `math.lua` rather than `objects.lua`. That'd more closely match where the object code lives in love's codebase and where the documentation for those objects lives in the wiki.

ellraiser:
Sure! Can do that - do you still want them accessible under the love.test.objects "module" or would you prefer having them directly under the module, i.e. love.test.graphics.Rasterizer()?

slime73:
Directly under the module like `love.test.graphics.Texture.getDimensions` probably makes sense, although that's a bit wordy

testing/tests/audio.lua
```diff
@@ -0,0 +1,446 @@
...
+love.test.audio.Source = function(test)
+  -- create stereo source
+  local stereo = love.audio.newSource('resources/click.ogg', 'static')
+  test:assertObject(stereo)
+  -- check stereo props
+  test:assertEquals(2, stereo:getChannelCount(), 'check stereo src')
+  test:assertEquals(66, math.floor(stereo:getDuration("seconds")*1000), 'check stereo seconds')
+  test:assertNotNil(stereo:getFreeBufferCount())
+  test:assertEquals('static', stereo:getType(), 'check stereo type')
+  -- check cloning a stereo
```
slime73, line 60: Can each section like this have an empty line before it? Things are super compressed in the test code right now, which makes it hard to read

testing/tests/physics.lua
```diff
@@ -0,0 +1,685 @@
...
+  shape1:setFriction(1)
+  test:assertEquals(1, shape1:getFriction(), 'check set friction')
+  test:assertEquals(0, shape1:getRestitution(), 'check def restitution')
+  shape1:setRestitution(0.5)
+  test:assertEquals(5, math.floor(shape1:getRestitution()*10), 'check set restitution')
+  -- check points
+  local shape2 = love.physics.newRectangleShape(body1, 5, 5, 10, 10)
+  tlx, tly, brx, bry = shape2:getBoundingBox(1)
+  test:assertEquals(true, shape2:testPoint(5, 5), 'check point 5,5')
+  test:assertEquals(true, shape2:testPoint(15, 15, 10, 10, 0), 'check point 15,15 after translate 10,10')
```
slime73, line 318: Aside from my recent testPoint return value fix, I think the parameters here might not be right. As per the wiki page, the parameter order is (translateX, translateY, transformAngleRadians, pointX, pointY)`. So it's checking point (10, 0) after translating by (15, 15) and rotating by 10 radians, which seems wrong.

testing/tests/physics.lua
```diff
@@ -0,0 +1,685 @@
...
+  test:assertEquals(111, math.floor(mass), 'check mass at 1000 density')
+  test:assertEquals(7407, math.floor(inertia), 'check intertia at 1000 density')
+  test:assertEquals(2, math.floor(shape1:getFriction()*10), 'check def friction')
+  shape1:setFriction(1)
+  test:assertEquals(1, shape1:getFriction(), 'check set friction')
+  test:assertEquals(0, shape1:getRestitution(), 'check def restitution')
+  shape1:setRestitution(0.5)
+  test:assertEquals(5, math.floor(shape1:getRestitution()*10), 'check set restitution')
+  -- check points
+  local shape2 = love.physics.newRectangleShape(body1, 5, 5, 10, 10)
```
slime73, line 315: Should this be reusing the dynamic body, considering it's testing a world point below?

testing/tests/physics.lua
```diff
@@ -0,0 +1,685 @@
...
+
+-- love.physics.getDistance
+love.test.physics.getDistance = function(test)
+  -- setup two fixtues to check
+  local world = love.physics.newWorld(0, 0, false)
+  local body = love.physics.newBody(world, 10, 10, 'static')
+  local shape1 = love.physics.newEdgeShape(body, 0, 0, 5, 5)
+  local shape2 = love.physics.newEdgeShape(body, 10, 10, 15, 15)
+  -- check distance between them
+  test:assertEquals(647106, math.floor(love.physics.getDistance(shape1, shape2)*100000), 'check distance matches')
```
slime73, line 471: Is there a sort of 'assert number equals with the given tolerance' helper function this could use, instead of having to apply ad hoc tolerance code?

testing/tests/physics.lua
```diff
@@ -0,0 +1,685 @@
...
+  test:assertEquals(0, math.floor(shape1:getRadius()), 'check radius')
+  test:assertEquals('polygon', shape1:getType(), 'check rectangle type')
+  test:assertEquals(0, shape1:getBody():getX(), 'check body link')
+  test:assertEquals(1, shape1:getCategory(), 'check def category')
+  shape1:setCategory(3, 5, 6)
+  local categories = {shape1:getCategory()}
+  test:assertEquals(14, categories[1] + categories[2] + categories[3], 'check set category')
+  test:assertEquals(false, shape1:isSensor(), 'check sensor def')
+  shape1:setSensor(true)
+  test:assertEquals(true, shape1:isSensor(), 'check set sensor')
```
slime73, line 277: `assertTrue` and `assertFalse` convenience methods might be handy

slime73 (changes requested):

testing/tests/physics.lua
pablomayobre: Also I recommend not repeating values when doing these sort of test cases, so that you can easily identify if for example X and Y are swapped for some reason

testing/main.lua
```diff
@@ -0,0 +1,216 @@
...
+-- run test suite logic 
+love.update = function(delta)
+  love.test:runSuite(delta)
+end
+
+
+-- love.draw
+-- draw a little logo to the screen
+love.draw = function()
+  local lw = (love.graphics.getPixelWidth() - 128) / 2
```
slime73, line 176: I think this will draw off-screen when high dpi is used - `love.graphics.getWidth` (and getHeight for the line below) should be used instead

testing/main.lua
```diff
@@ -0,0 +1,216 @@
...
+
+  -- convert args to the cmd to run, modules, method (if any) and disabled
+  local testcmd = '--runAllTests'
+  local module = ''
+  local method = ''
+  local cmderr = 'Invalid flag used'
+  local modules = {
+    'audio', 'data', 'event', 'filesystem', 'font', 'graphics',
+    'image', 'math', 'physics', 'sound', 'system',
+    'thread', 'timer', 'video', 'window'
```
slime73, line 78: love 12 has a new `love.sensor` module. It's all about inputs from hardware sensors so testing actual values won't work probably, but we shouldn't forget about the basics

testing/main.lua
```diff
@@ -0,0 +1,216 @@
...
+      love.graphics.setLineStyle('rough')
+      love.graphics.setLineWidth(1)
+      Logo = {
+        texture = love.graphics.newImage('resources/love.png'),
+        img = nil
+      }
+      Logo.img = love.graphics.newQuad(0, 0, 64, 64, Logo.texture)
+      Font = love.graphics.newFont('resources/font.ttf', 8, 'normal')
+      local txtobj = love.graphics.newTextBatch or love.graphics.newText
+      TextCommand = txtobj(Font, 'Loading...')
```
slime73, line 51: I think these can just be `love.graphics.print` calls inside love.draw instead of needing TextBatch objects

testing/classes/TestSuite.lua
```diff
@@ -0,0 +1,183 @@
...
+  -- @method - TestSuite:runSuite()
+  -- @desc - called in love.update, runs through every method or every module
+  -- @param {number} delta - delta from love.update to track time elapsed
+  -- @return {nil}
+  runSuite = function(self, delta)
+
+      -- stagger between tests
+    if self.module ~= nil then
+
+        if self.module.start == true then
```
slime73, line 63: `if foo == true then` can be simplified to `if foo then` (and `if foo == false then` can be `if not foo then`)

slime73 (approved):
I left a few comments but I don't think any of them need to be addressed before this version is merged. This was a huge amount of effort, thanks so much for taking it on! I like that it's already found several legit bugs in love 12 even while it was being fleshed out.

slime73:
I might have more comments once I start adding to the tests myself, but for now I'll press the big green button and see what happens

PR #1985, opened 2023-11-24
MinGW / MSYS2 Support

sewbacca:
This PR enables support for compiling love2d under MinGW/MSYS2.

Tested on msys2/ucrt64 (Windows 10 64 bit).
All build dependencies are available via pacman (msys2).


CMakeLists.txt
```diff
@@ -179,7 +179,7 @@ if(MEGA)
 else()
 	if(MSVC)
 		message(FATAL_ERROR "
-It is currently only possible to build with megasource on Windows.
+It is currently only possible to build with megasource or mingw on Windows.
```
slime73, line 182: Could we add a notice or warning while using mingw that it's not officially supported? Since we don't plan to maintain it ourselves, so it might break in the future.

src/love.cpp
```diff
@@ -292,4 +292,11 @@ int main(int argc, char **argv)
 	return retval;
 }
 
+#ifdef __MINGW32__
+int WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int nShowCmd)
```
slime73, line 296: Isn't the SDL2main library supposed to handle this?

CMakeLists.txt
sewbacca: Sure that makes sense, adjusted in 88ea145.

src/love.cpp
sewbacca: I didn't know this, fixed in df108cb.

CMakeLists.txt
```diff
@@ -1790,11 +1799,16 @@ add_executable(${LOVE_EXE_NAME} WIN32 src/love.cpp ${LOVE_RC})
...
 set_target_properties(${LOVE_EXE_NAME} PROPERTIES C_VISIBILITY_PRESET hidden CXX_VISIBILITY_PRESET hidden VISIBILITY_INLINES_HIDDEN ON)
 
-if(MSVC)
+if(MSVC OR MINGW)
 	add_executable(${LOVE_CONSOLE_EXE_NAME} src/love.cpp ${LOVE_RC})
 	target_link_libraries(${LOVE_CONSOLE_EXE_NAME} ${LOVE_LIB_NAME})
 endif()
 
+if(MINGW)
+	target_compile_definitions(${LOVE_EXE_NAME} PRIVATE SDL_MAIN_HANDLED)
```
slime73, line 1808: Is this actually needed? I thought it was only for situations where the app definitely doesn't want to use SDL's main entry point redirection (whereas love should use it). I think it also needs an extra function call in the main entry point if it's defined, too.

sewbacca: You are right, I misunderstood the definition. I thought it would enable the SDL2 main, when in fact, it did quite the opposite. The reason I put it there was because of an undefined reference to WinMain. After [some digging](https://wiki.libsdl.org/SDL2/FAQWindows#i_get_undefined_reference_to_winmain16), I found out, that I need to link against `mingw32` and `SDL2main` before linking against `SDL2`, as you have mentioned earlier. It's now fixed in [b08aa24](
https://github.com/love2d/love/commit/b08aa244c5d5c03c3a468e77dcd58d20fcb77475#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR225-R231).

CMakeLists.txt
```diff
@@ -222,6 +222,14 @@ Please see https://github.com/love2d/megasource
 		${ZLIB_LIBRARY}
 	)
 
+	if(MINGW)
+		set(LOVE_LINK_LIBRARIES
+			mingw32
+			SDL2main
```
slime73, line 228: I was curious why it wasn't already linking with SDL2main so I did some googling (more than I would have liked to do), and I think it might work to change the include dirs and link libraries above this from `SDL2_INCLUDE_DIR` and `SDL2_LIBRARY` to `SDL2_INCLUDE_DIRS` and `SDL2_LIBRARIES`, in order to link with SDL2main on windows?

If that works I think it'd be good to change this to do that instead, since it's cleaner.

sewbacca: That does work, however one needs to configure `find_package` to look for the `SDL2main`. I oriented myself by [SDL2's README for cmake](https://wiki.libsdl.org/SDL2/README/cmake#including-sdl-in-your-project). Since the config has no `REQUIRED` attribute, it should not break anything on other plattforms, I suppose.
Adjusted in 11d4bdd.

slime73:
Looks good to me!

PR #1989, opened 2023-12-04
test suite cleanup + updates

ellraiser:
- update main love readme to include note on tests
- added module tests for Sensor, Keyboard, Touch, Joystick, Mouse, and Love (for hooks + basic love-level methods)
- general clean-up of main class logic per initial merge comments
- couple fixes for runner-specific failures (physics.Joint, audio.getActiveSourceCount, mouse.getPosition)
- change window.close test to run last


ellraiser:
have seen that video failure crop up a couple times now on runners randomly - will put something in to handle

testing/classes/TestMethod.lua
```diff
@@ -379,17 +379,17 @@ TestMethod = {
     local failure = ''
     local failures = 0
     for a=1,#self.asserts do
-      -- @TODO just return first failed assertion msg? or all?
+      -- @TODO show all failued assertion methods?
```
slime73, line 382: failued -> failed

testing/tests/joystick.lua
```diff
@@ -0,0 +1,81 @@
...
+love.test.joystick.getJoystickCount = function(test)
+  local count = love.joystick.getJoystickCount()
+  test:assertEquals(0, count, 'check no joysticks')
+end
+
+
+-- love.joystick.getJoysticks
+love.test.joystick.getJoysticks = function(test)
+  local joysticks = love.joystick.getJoysticks()
+  test:assertEquals(0, #joysticks, 'check no joysticks')
```
slime73, line 29: I don't think we should check the count here, since I'd expect tests to pass on my computer with an xbox controller connected and with it not connected.

testing/tests/keyboard.lua
```diff
@@ -0,0 +1,73 @@
...
+--------------------------------------------------------------------------------
+------------------------------------METHODS-------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+
+-- love.keyboard.getKeyFromScancode
+love.test.keyboard.getKeyFromScancode = function(test)
+  local key = love.keyboard.getKeyFromScancode('a')
+  test:assertEquals('a', key, 'check key')
```
slime73, line 15: This is dependent on the physical keyboard and system configuration for it, so it shouldn't check for a specific value.

testing/tests/keyboard.lua
```diff
@@ -0,0 +1,73 @@
...
+end
+
+
+-- love.keyboard.getScancodeFromKey
+love.test.keyboard.getScancodeFromKey = function(test)
+  local scancode = love.keyboard.getScancodeFromKey('a')
+  test:assertEquals('a', scancode, 'check scancode')
```
slime73, line 22: Same here

testing/tests/love.lua
```diff
@@ -0,0 +1,78 @@
...
+end
+
+
+-- love.errhand
+love.test.love.errhand = function(test)
+  test:assertTrue(type(love.errhand) == 'function', 'check defined')
+end
+
+
+-- love.load
```
slime73, line 51: We probably don't need to check load/draw/update because they're defined by user code (this code), not by love.

testing/tests/mouse.lua
```diff
@@ -0,0 +1,173 @@
...
+------------------------------------METHODS-------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+
+-- love.mouse.getCursor
+love.test.mouse.getCursor = function(test)
+  local cursor = love.mouse.getCursor()
+  test:assertEquals(nil, cursor, 'check nil initially')
+  love.mouse.setCursor(love.mouse.getSystemCursor("hand"))
```
slime73, line 16: This test might fail on mobile systems that don't have cursors

testing/tests/mouse.lua
```diff
@@ -0,0 +1,173 @@
...
+  test:assertEquals(true, visible, 'check visible default')
+  love.mouse.setVisible(false)
+  test:assertEquals(false, love.mouse.isVisible(), 'check invisible')
+  love.mouse.setVisible(true)
+end
+
+
+-- love.mouse.newCursor
+love.test.mouse.newCursor = function(test)
+  local cursor = love.mouse.newCursor('resources/love.png', 0, 0)
```
slime73, line 100: I'm not sure if newCursor works on systems that don't support cursors

testing/tests/physics.lua
```diff
@@ -307,10 +307,10 @@ love.test.physics.Joint = function(test)
   test:assertFalse(joint:isDestroyed(), 'check not destroyed')
 
   -- check reaction props (sometimes nil on linux runners)
-  if joint:getReactionForce(1) ~= nil then
+  if joint:getReactionForce(1) ~= nil and joint:getReactionForce(1) ~= 0/0 then
```
slime73, line 310: Do we know why this is failing sometimes? I think if there's a random failure on CI machines like this I'd rather either investigate it or have this part of the test very obviously temporarily turned off until it can be investigated.

testing/tests/touch.lua
```diff
@@ -0,0 +1,33 @@
...
+  test:assertNotNil(love.touch.getPressure)
+  test:assertEquals('function', type(love.touch.getPressure))
+end
+
+
+-- love.touch.getTouches
+love.test.touch.getTouches = function(test)
+  local touches = love.touch.getTouches()
+  test:assertNotNil(touches)
+  test:assertEquals(0, #touches, 'check no touches')
```
slime73, line 32: I don't think we can rely on the count while tests are running

slime73 (changes requested):

testing/tests/physics.lua
```diff
@@ -307,10 +307,10 @@ love.test.physics.Joint = function(test)
...
     test:assertEquals(0, joint:getReactionForce(1), 'check reaction force')
```
slime73, line 311: getReactionForce also returns 2 values

testing/tests/physics.lua
slime73: Answer: I can't find any situation where getReactionForce is able to return nil in the code, but the numbers it does return (as well as possible other Joint getters) are not valid until the World has been updated at least once after the Joint is created, I think.

ellraiser: yeah ideally I'd like to leave if it's a failure but it's been so random, I'd say maybe 1/20 of the workflow actions I've run I've seen it popup

I'll try doing a world update before those joint calls, and remove the ~= nil ~= 0/0 checks to let it fail and we can see if having the world update solves it happening - if not can decide what you wanna do

testing/tests/joystick.lua
slime73: Maybe change it to assertGreaterEqual like the getJoystickCount test above?

slime73 (approved):

Issue #1992, opened 2023-12-11
[Love 12.0][Feature request] Ability to provide debug name for buffer/texture resources generated by Love


Froyok:
Hi,

When using a graphic debugger (like [RenderDoc](https://renderdoc.org/)) to inspect a frame made with Löve, the canvas and buffers generated by Löve have generic names (ex: `Texture 471`). 

This makes things a bit difficult sometimes to track bugs or even resources across multiple drawcalls.

I imagine that currently those names are automatic, defined by the graphic API when it creates the objects. It seems however that debug names can be provided, which could help identify more easily those resources: https://renderdoc.org/docs/how/how_annotate_capture.html#application-provided-object-names

Would it be possible to expose a way to specify this debug name on the Lua side ?
Maybe something like `Texture:setDebugName()` and `Texture:getDebugName()`.

slime73:
This is implemented now. love's built-in/default textures and buffers don't have unique names yet, but that's partly because they aren't represented by actual Texture and Buffer objects on all backends. I already want to clean that up for unrelated reasons, so I'll add debug names to those when I do that.

Froyok:
Looking at the review (haven't tried it yet), and not being familiar enough with the sources, does it works as well with `love.graphics.newImage` for example ? Or is it only canvas/buffers ?

nikeinikei:
It works for any texture, image and canvas got merged into one in Löve 12.

Froyok:
Great, thank you for confirming. :)

Froyok:
Can confirm, it works really well ! Thank you both ! :heart: 
![2023-12-14__20-28-59](https://github.com/love2d/love/assets/6290387/c6b194b1-fdc3-44da-becb-80213e745a53)

PR #1994, opened 2023-12-13
Implement #1992

nikeinikei:
The settings table for Texture and GraphicsBuffer can now include a new field `debugname`,
If given, it will appear in a graphics debugger such as RenderDoc, if supported.
it can also be queried via `texture:getDebugName()` or `buffer:getDebugName()`, which either returns a `string` if a debug name was given during initialization, or `nil` otherwise.

Example:
```lua
local texture = love.graphics.newCanvas(64, 64, { debugname = "shadowpass" })
print(texture:getDebugName())    --prints "shadowpass"

local format = {
    { name = 'position', format = 'floatvec2' },
    { name = 'velocity', format = 'floatvec2' },
    { name = 'color', format = 'floatvec4' }
}
local count = 1000000

local buffer = love.graphics.newBuffer(format, count, { shaderstorage = true, debugname = "particles" })
print(buffer:getDebugName())    --prints "particles"
```

### DIFF
--- a/src/modules/graphics/Texture.h
+++ b/src/modules/graphics/Texture.h
@@ -195,7 +195,7 @@ public:
 		bool renderTarget = false;
 		bool computeWrite = false;
 		OptionalBool readable;
-		Optional<std::string> debugName;
+		std::string debugName;
 	};
 
 	struct Slices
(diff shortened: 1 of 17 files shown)


src/modules/graphics/Texture.h
```diff
@@ -194,6 +195,7 @@ class Texture : public Drawable, public Resource
 		bool renderTarget = false;
 		bool computeWrite = false;
 		OptionalBool readable;
+		Optional<std::string> debugName;
```
slime73, line 198: This could be a `std::string` without the Optional wrapper, and then the backend code could check whether the string is empty before calling the graphics API to set it.

PR #1999, opened 2023-12-24
Fix deprecated love.graphics.stencil function resetting stencilMode

Raltyro:
Quite self-explanatory
Games from previous versions using deprecated stencil functions will most likely not working as intended

```lua
love.graphics.stencil(stencil, "replace", 1, false)
love.graphics.setStencilTest("greater", 0)
```
Example above vice versa would work the same in previous love2d version games but since currently stencil function resets the properties it wouldnt work the same

This pr fixes it


slime73:
Nice catch, thanks!

PR #2001, opened 2023-12-31
Replace autotools with cmake

### LINKED ISSUE #1991: Phase out Autotools Build System
During the release of 11.5, we have bit of discussion on switching the Linux build system from Autotools to CMake. This means one build system for every platform we support (except macOS families), although Linux enthusiast users may not happy with this change.

I left this open for further discussion ~~and didn't tag it to specific milestones for same reason~~.

bartbes:
Resolves #1991


CMakeLists.txt
```diff
@@ -1996,6 +2018,10 @@ add_library(${LOVE_LIB_NAME} SHARED ${LOVE_LIB_SRC} ${LOVE_RC})
 set_target_properties(${LOVE_LIB_NAME} PROPERTIES C_VISIBILITY_PRESET hidden CXX_VISIBILITY_PRESET hidden VISIBILITY_INLINES_HIDDEN ON)
 target_link_libraries(${LOVE_LIB_NAME} ${LOVE_LINK_LIBRARIES} ${LOVE_3P})
 
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+	set_target_properties(${LOVE_LIB_NAME} PROPERTIES LIBRARY_OUTPUT_NAME "${LOVE_EXE_NAME}-${LOVE_VERSION_STR}")
```
bartbes, line 2022: This also seems vaguely related to the `set(LOVE_LIB_NAME)` for android. Maybe `LOVE_EXE_NAME` and `LOVE_LIB_NAME` should just set the `OUTPUT_NAME` properties, rather than actually being the target names?

CMakeLists.txt
```diff
@@ -2042,142 +2068,140 @@ if(NOT ANDROID)
...
+	include(GNUInstallDirs)
+
+	file(COPY platform/unix/debian/ DESTINATION debian PATTERN "*.in" EXCLUDE)
+	configure_file(platform/unix/debian/control.in debian/control @ONLY)
+	configure_file(platform/unix/debian/changelog.in debian/changelog @ONLY)
+	configure_file(platform/unix/debian/rules.in debian/rules @ONLY)
+
+	configure_file(platform/unix/love.desktop.in love.desktop @ONLY)
+
+	install(TARGETS ${LOVE_EXE_NAME} ${LOVE_LIB_NAME})
```
bartbes, line 2195: Autotools also installs a `liblove.so` symlink. As nothing explicitly links against it (except user code?) I'm don't mind skipping that.

CMakeLists.txt
```diff
@@ -2042,142 +2068,140 @@ if(NOT ANDROID)
...
+		DeleteRegKey HKCR \\\".love\\\"
+		System::Call 'shell32.dll::SHChangeNotify(i, i, i, i) v  (0x08000000, 0, 0, 0)'
+	")
+
+	include(CPack)
+elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+	include(GNUInstallDirs)
+
+	file(COPY platform/unix/debian/ DESTINATION debian PATTERN "*.in" EXCLUDE)
+	configure_file(platform/unix/debian/control.in debian/control @ONLY)
```
bartbes, line 2189: Of course I realise _now_ that I only checked whether there was something sane for the no-suffix case, and never checked if the suffix stuff worked. That probably needs to be fixed.

CMakeLists.txt
slime73: Would that be for allowing `require("love")` to work from a standalone Lua executable?

bartbes: Sort of, it's still not in the normal lua search paths.

siddle1512:
nhu cac

CMakeLists.txt
MikuAuahDark: They'll have to do `package.preload.love = package.loadlib("path/to/liblove-12.0.so", "luaopen_love")` before doing `require("love")` from my understanding, at least.

CMakeLists.txt
MikuAuahDark: Good idea, although this need extensive testing both on Windows and Android.

PR #2003, opened 2024-01-09
Emit love.visible(true) on SDL_WINDOWEVENT_EXPOSED

doesntgolf:
I'm using the i3 window manager, and SDL is firing the `SDL_WINDOWEVENT_EXPOSED` when I switch from one workspace to the workspace Love is on. The SDL wiki [says](https://wiki.libsdl.org/SDL2/SDL_WindowEventID) `SDL_WINDOWEVENT_EXPOSED` means the "window has been exposed and should be redrawn". I figured this fits in well enough to `love.visible` rather than needing its own callback.

### DIFF
--- a/src/modules/event/sdl/Event.cpp
+++ b/src/modules/event/sdl/Event.cpp
@@ -603,7 +603,8 @@ Message *Event::convertWindowEvent(const SDL_Event &e)
 		break;
 	case SDL_WINDOWEVENT_SHOWN:
 	case SDL_WINDOWEVENT_HIDDEN:
-		vargs.emplace_back(e.window.event == SDL_WINDOWEVENT_SHOWN);
+	case SDL_WINDOWEVENT_EXPOSED:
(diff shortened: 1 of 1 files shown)


slime73:
`exposed` has a slightly different meaning than becoming visible - it basically means something has happened to make the contents of the window no longer valid (for example the window manager decided to trash the backbuffer's data for whatever reason).

So I'd rather not send a visible event when it happens because the SDL event isn't saying the window transitioned from non-visible to visible.

slime73:
SDL3 introduces a new `occluded` event which says when the window exists but is no longer showing up - for example it's covered by other opaque windows or it's in a different virtual workspace from the user's active one. So it's kind of the other half of the goals of this change.

I think I'd rather wait for SDL3 to stabilize and love to switch to it before introducing something like this.

slime73:
If you want to open a new issue in the issue tracker that might be handy, so we don't forget about the feature (and the reasons for wanting it).

Issue #2015, opened 2024-02-06
love.data.hash container output typed for love 12.0


EngineerSmith:
To match other functions within the `love.data` module. I'll like to see `love.data.hash` accept an enum for if it should return `Data` or a `String`. Right now, it only return a string with no option for data.

This will be a breaking change, but it "shouldn't be too difficult"™ to add.

Why should this be implemented? Well, having to create a separate bytedata when I want it in data form isn't easy to make the code readable. Plus, it doesn't following the other conventions of `love.data`. Considering that the returned value cannot be logged, or printed without being encoded to base64 - doesn't really suit being stored as a string.

If you wanted to save it? File supports writing data. 
If you wanted to send it across threads? Variant supports data.
If you wanted to log it? Encoder supported data input

I have no idea why you'd want it to be a string, but it should still have the option to be a string.

```lua
ld = love.data
data = love.data.newByteData(...)

hashed = ld.newByteData(ld.hash("sha512", data))
-- vs
hashed = ld.hash("data", "sha512", data)

-- and it's compliment
hashed = ld.hash("string", "sha512", data)
```

https://github.com/love2d/love/blob/12.0-development/src/modules/data/wrap_DataModule.cpp#L323

slime73:
Addressed in [483237e](https://github.com/love2d/love/commit/483237e).

PR #2018, opened 2024-02-09
Added container type argument for love.data.hash

EngineerSmith:
Changed `love.data.hash` to accept a container type as an argument. This will be a breaking change!

I made the issue #2015, and decided to implement it. If you believe this change is too breaking, or unnecessary, feel free to reject it.

This PR also updates the tests, and adds the change to the `changes.txt`
```lua
local ld = love.data
local data = ld.newByteData("helloworld")

-- old
local hashed = ld.hash("sha512", data)
-- above replaced by
local hashed = ld.hash("string", "sha512", data)

-- new functionality
local hashed = ld.hash("data", "sha512", data)
```

### DIFF
--- a/src/modules/data/wrap_DataModule.cpp
+++ b/src/modules/data/wrap_DataModule.cpp
@@ -302,25 +302,37 @@ int w_decode(lua_State *L)
 
 int w_hash(lua_State *L)
 {
-	const char *fstr = luaL_checkstring(L, 1);
+	ContainerType ctype = luax_checkcontainertype(L, 1);
+	const char *fstr = luaL_checkstring(L, 2);
 	HashFunction::Function function;
 	if (!HashFunction::getConstant(fstr, function))
 		return luax_enumerror(L, "hash function", HashFunction::getConstants(function), fstr);
 
 	HashFunction::Value hashvalue;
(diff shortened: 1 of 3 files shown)


src/modules/data/wrap_DataModule.cpp
```diff
@@ -302,25 +302,37 @@ int w_decode(lua_State *L)
 
 int w_hash(lua_State *L)
 {
-	const char *fstr = luaL_checkstring(L, 1);
+	ContainerType ctype = luax_checkcontainertype(L, 1);
```
slime73, line 305: Could we keep the original two-parameter variant but mark it as deprecated (with love's deprecation APIs)?

slime73 (changes requested):
The `12.0-development` branch has been merged into `main` now, you'll need to change the base branch of this PR to `main` as well.

src/modules/data/wrap_DataModule.cpp
EngineerSmith: This has been resolved. The image shows the deprecated message, along with checking that it returns string type as it did in the previous version without the container type defined. image

slime73 (approved):
Looks good now, thanks!

PR #2021, opened 2024-02-11
typos

a-detiste:


slime73:
These changes are all in third party libraries - I'd rather not modify them locally like this for reasons other than to fix breaking bugs. You could submit PRs to their original repositories though (assuming those typos haven't already been fixed in later versions).

a-detiste:
OK, I'll look. I uploaded 11.5 today to Debian (with liblove staticaly linked)

a-detiste:
VMA is already fixed

PR #2026, opened 2024-02-26
Correct compilation instructions for *nix

akkartik:
It looks like the old `automagic` file is no longer in the repo.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -54,7 +54,7 @@ Compilation
 Follow the instructions at the [megasource][megasource] repository page.
 
 ### *nix
-Pick a directory to store generated files to, say `build/`:
+First pick a directory to store generated files to, say `build/` and then:
 
 	$ cmake -B build -S. --install-prefix $PWD/prefix # this will create the directory `build/`
 	$ cmake --build build --target install # this will put the files in `prefix/`.
--- a/readme.md
+++ b/readme.md
@@ -56,9 +56,8 @@ Follow the instructions at the [megasource][megasource] repository page.
 ### *nix
 Pick a directory to store generated files to, say `build/`:
 
-	$ cmake -B build  # this will create the directory `build/`
-	$ cd build/
-	$ make
+	$ cmake -B build -S. --install-prefix $PWD/prefix # this will create the directory `build/`
+	$ cmake --build build --target install # this will put the files in `prefix/`.
 
 ### macOS
 Download or clone [this repository][dependencies-apple] and copy, move, or symlink the `macOS/Frameworks` subfolder into love's `platform/xcode/macosx` folder.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -54,10 +54,10 @@ Compilation
...
 
 ### *nix
-Run `platform/unix/automagic` from the repository root, then run ./configure and make.
+Pick a directory to store generated files to, say `build/`:
 
-	$ platform/unix/automagic
-	$ ./configure
+	$ cmake -B build  # this will create the directory `build/`
+	$ cd build/
 	$ make
```
MikuAuahDark, line 61: I suggest not to invoke `make` directly as this is not portable.
```suggestion
	$ cmake -B build -S. --install-prefix $PWD/prefix # this will create the directory `build/`
	$ cmake --build build --target install # this will put the files in `prefix/`.
```

MikuAuahDark (commented):
The wordings before the command probably can be made better.

akkartik:
How does this look?

MikuAuahDark:
I'm still not too positive with the wording at "First pick a directory to store generated files to, say `build/` and then:". Probably that's just me. Perhaps anyone else have better opinion?

akkartik:
Yeah, I'm not attached to it. Feel free to suggest something. Do you want it to be more explicit? Basically my thinking/preference was the commands are already saying what to do. But happy to add whatever you like.

readme.md
```diff
@@ -54,13 +54,10 @@ Compilation
...
-Run `platform/unix/automagic` from the repository root, then run ./configure and make.
+First pick a directory to store generated files to, say `build/` and then:
 
-	$ platform/unix/automagic
-	$ ./configure
-	$ make
-
-When using a source release, automagic has already been run, and the first step can be skipped.
+	$ cmake -B build -S. --install-prefix $PWD/prefix # this will create the directory `build/`
+	$ cmake --build build --target install # this will put the files in `prefix/`.
```
ysharlaimov, line 60: `build` and `prefix` folders will popup as untracked files in git repository. Is it ok ?

ysharlaimov: `cmake -S.` also implies that we run these commands from the repository root. Probably we should explicitly mention it

ysharlaimov: As I see in CMakeLists.txt, you can't run `cmake -B. -S.`, but it should be ok for `cmake -Bbuild -S.`

readme.md
```diff
@@ -54,13 +54,10 @@ Compilation
 Follow the instructions at the [megasource][megasource] repository page.
 
 ### *nix
-Run `platform/unix/automagic` from the repository root, then run ./configure and make.
+First pick a directory to store generated files to, say `build/` and then:
```
learn-more, line 57:
```suggestion
Because in-tree builds are not allowed, you have to create Makefiles in a build folder.
In this example we will use the `build/` folder:
```

MikuAuahDark:
I opted slightly different text based on this PR. Thanks.

PR #2036, opened 2024-03-10
Modernize cmake builds

bartbes:
Here are some smaller and larger changes to CMakeLists.txt I thought of when converting the linux builds to cmake.

You're definitely welcome to pick and choose parts of this PR you'd like to merge (if any). I think all changes are net improvements, but I realize some might be controversial, especially the modules-as-targets one, with its extra dependency tracking.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,7 +247,7 @@ endfunction()
 # common
 #
 
-set(LOVE_SRC_COMMON
+add_library(love_common STATIC
 	src/common/android.cpp
 	src/common/android.h
 	src/common/b64.cpp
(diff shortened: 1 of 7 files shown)


src/modules/audio/openal/Audio.h
```diff
@@ -50,9 +50,9 @@
 #include <OpenAL-Soft/al.h>
 #endif
 #else
-#include <AL/alc.h>
-#include <AL/al.h>
-#include <AL/alext.h>
+#include <alc.h>
+#include <al.h>
+#include <alext.h>
 #endif
```
bartbes, line 56: I wonder if the shortened include paths mean the ifdefs for macos/iOS can also be cleaned up, presumably they can now set the full include path in their project files.

slime73: I think macOS frameworks (via the `-framework` parameter rather than `-l` or whatever) still require the framework name in the include path, and love's xcode project still uses frameworks.

slime73:
Every part of this looks good to me, but I don't know cmake very well so maybe someone else can weigh in.

Is there any reason not to remove the draft tag from this and merge the whole thing?

learn-more:
> Every part of this looks good to me, but I don't know cmake very well so maybe someone else can weigh in.
> 
> Is there any reason not to remove the draft tag from this and merge the whole thing?

These changes make sense from a CMake perspective, the `interface` libraries are a good way to propagate include directories, defines and libraries.

One thing to note is the minimum cmake version bump to `3.18`, that is released on `July 15, 2020`.
I am not sure if all (or most common) distributions have that version in their package managers.

slime73:
SDL3 has its minimum cmake version set to 3.16, for a reference point.

CMakeLists.txt
```diff
@@ -23,18 +23,13 @@ if(${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_BINARY_DIR})
 	message(FATAL_ERROR "Prevented in-tree build.")
 endif()
 
-cmake_minimum_required(VERSION 3.7)
+cmake_minimum_required(VERSION 3.18)
```
MikuAuahDark, line 26: I think CMake 3.18 is bit too high but I don't think it's an issue for us. It could be an issue with downstream though.

MikuAuahDark (approved):
In overall, it looks good to me. Android needs an additional patch in its Java side to load `libliblove.so` in addition to `liblove.so` but I can do that once this is merged. 

Feel free to switch it to complete PR.

bartbes:
I picked 3.18 because that's what debian bullseye (oldstable) has, I can downgrade it if you want.

slime73:
If 3.18 isn't actively needed then I think using 3.16 instead (for now at least) makes sense to me, but I don't really have a strong opinion.

bartbes:
Oops, removed the wrong branch.

PR #2037, opened 2024-03-11
vk: unify local uniform buffer

nikeinikei:


src/modules/graphics/vulkan/Graphics.cpp
```diff
@@ -2869,6 +2870,23 @@ int Graphics::getVsync() const
 	return vsync;
 }
 
+void Graphics::mapLocalUniformData(void *data, size_t size, VkDescriptorBufferInfo &bufferInfo)
+{
+	size_t alignedSize = static_cast<size_t>(std::ceil(static_cast<float>(size) / static_cast<float>(minUniformBufferOffsetAlignment))) * minUniformBufferOffsetAlignment;
```
slime73, line 2875: I think this can use love's `alignUp` function

nikeinikei: ah true, that makes the code a lot nicer

PR #2039, opened 2024-03-16
Android Test Suites

MikuAuahDark:
The test suites assume certain environment condition which can't be satisfied in Android. In particular:
* `love.filesystem.getSource()` does not point to writeable directory. This is true if the test suite program is packaged as .love file or when fused. This is the case for Android test suite (in the future).
* Based on `love.filesystem.getSource()` quirks, then the `mountFullPath` and `unmountFullPath` tests are skipped entirely.
* `love.window.getMode()` always return near-fullscreen dimensions. In this case, skip the width/height checks.
* Maximizing window is not supported.
* Minimizing window works, **but pauses** the test suite execution, so skip this too.

Marking it as draft as I haven't wrote the actual automated Android testing using AVD yet.


testing/tests/filesystem.lua
```diff
@@ -393,7 +403,7 @@ love.test.filesystem.mountCommonPath = function(test)
   local mount3 = love.filesystem.mountCommonPath('userhome', 'userhome', 'readwrite')
   local mount4 = love.filesystem.mountCommonPath('userappdata', 'userappdata', 'readwrite')
   -- userdesktop isnt valid on linux
-  if love.system.getOS() ~= 'Linux' then
+  if love._os ~= 'Linux' and love._os ~= 'Android' then
```
slime73, line 406: I think these can still use `love.system.getOS()`

MikuAuahDark: Wasn't sure about that due to this notice in tests main.lua
https://github.com/love2d/love/blob/a3c84ba409405a8648bffc5381f2aad5e63139ba/testing/main.lua#L26

slime73: I think we can consider love.system less of an optional module (especially for the test app), more like love.filesystem, so it should be fine to keep using it.

slime73: That being said, maybe having an `isOS(...)` function in the test framework itself could be handy so you don't have to write down multiple string comparisons in different places in the code.

PR #2041, opened 2024-03-19
test suite updates

ellraiser:
- changed tests so that all assert failures are logged rather than just the first assert that failed
- added isOS() to the test method class to be used in tests directly and replaced current love.system.getOS calls
- updated testsuite readme on a couple bits
- added newTexture basic check

### DIFF
--- a/testing/classes/TestMethod.lua
+++ b/testing/classes/TestMethod.lua
@@ -376,17 +376,14 @@ TestMethod = {
 
   -- @method - TestMethod:isOS()
   -- @desc - checks for a specific OS (or list of OSs)
-  -- @param {string||table} - single OS string or list of OSs that are allowed, 
-  --          these will be checked agaisnt love.system.getOS()'s return value
+  -- @param {string/s} - each arg passed will be checked as a valid OS, as long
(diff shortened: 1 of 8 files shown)


testing/classes/TestMethod.lua
```diff
@@ -374,33 +374,55 @@ TestMethod = {
   end,
 
 
+  -- @method - TestMethod:isOS()
+  -- @desc - checks for a specific OS (or list of OSs)
+  -- @param {string||table} - single OS string or list of OSs that are allowed, 
+  --          these will be checked agaisnt love.system.getOS()'s return value
+  -- @return {boolean} - returns true if one of the OSs given matches actual OS
+  isOS = function(self, oss)
```
slime73, line 382: This should use Lua varargs (`...`) instead of a table.

e.g.:

```Lua
isOS = function(self, ...)
  for i = 1, select("#", ...) do
    if love.test.current_os == select(i, ...) then
      return true
    end
  end
  return false
end
```

slime73 (changes requested):

slime73 (approved):

PR #2042, opened 2024-03-21
Make Graphics constructor explicit

zo9999:
This prevents unexpected implicit constructions of Graphics objects from C strings, since the constructor has C string as its single input parameter.

### DIFF
--- a/src/modules/graphics/Graphics.h
+++ b/src/modules/graphics/Graphics.h
@@ -451,7 +451,7 @@ public:
 		}
 	};
 
-	Graphics(const char *name);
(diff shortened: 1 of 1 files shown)


slime73:
I'm a little rusty on C++ implicit conversions, can you give a code example of where that could happen with the Graphics base class?

slime73:
`Graphics` is not a constructible object on its own (it's an abstract class, you'll get compile errors if you try to construct it) - so I don't think this change is relevant.

Also you're generally meant to construct love objects on the heap, and for modules in particular you shouldn't be constructing them yourself at all.

PR #2044, opened 2024-03-29
Add a workaround for Windows's initial window size

### LINKED ISSUE #1714: Window size incorrect when launching as maximized in Windows
https://github.com/26F-Studio/Techmino/issues/100

The above issue is on a game built using the Love engine. In summary, that person created a shortcut to the game executable, in right click -> properties -> shortcut, set "Run:" to Maximized, then opened the shortcut. When the game launched, the game's window is slightly larger than desirable, with a few pixels on the bottom of the window covered by the task bar.

I tried doing the same thing, but with the Love executable itself (without a game project), and it seems like this problem exists too, thus I believe this is a problem with the Love engine rather than the game.

We are both running Windows 10. There are screenshots in the issue linked above.

learn-more:
Fixes #1714

This completely ignores the window state requested in a shortcut or in a call to `CreateProcess`

### DIFF
--- a/src/love.cpp
+++ b/src/love.cpp
@@ -299,6 +299,35 @@ static DoneAction runlove(int argc, char **argv, int &retval, love::Variant &res
 	return done;
 }
 
+#ifdef LOVE_WINDOWS
+static INT_PTR CALLBACK EmptyDlgProc(HWND, UINT, WPARAM, LPARAM)
+{
+    return 0;
+}
+
+// Windows stores an initial window state passed in by the shell (or another process) if STARTF_USESHOWWINDOW is set.
+// This state is applied to the first window that receives a 'ShowWindow' call.
+// To ensure our window is not influenced by this, we create a dummy window to catch this initial state.
+// Ideally we would use SDL for this, but it skips calls to 'ShowWindow' if the window is hidden...
+static void love_windows_ignore_initial_wShowWindow()
+{
+    // Build up a dialog template
+    enum { DlgX = 0, DlgY = 0, DlgW = 100, DlgH = 60, ChildCount = 0, MenuId = 0, ClassName = 0, Title = 0 };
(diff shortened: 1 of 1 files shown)


src/love.cpp
```diff
@@ -299,6 +299,35 @@ static DoneAction runlove(int argc, char **argv, int &retval, love::Variant &res
...
+static INT_PTR CALLBACK EmptyDlgProc(HWND, UINT, WPARAM, LPARAM)
+{
+    return 0;
+}
+
+// Windows stores an initial window state passed in by the shell (or another process) if STARTF_USESHOWWINDOW is set.
+// This state is applied to the first window that receives a 'ShowWindow' call.
+// To ensure our window is not influenced by this, we create a dummy window to catch this initial state.
+// Ideally we would use SDL for this, but it skips calls to 'ShowWindow' if the window is hidden...
+static void love_windows_ignore_initial_wShowWindow()
```
slime73, line 312: Usually we don't do this sort of platform-specific workaround and leave it up to SDL instead - this sort of thing can be difficult for us to maintain and might have unexpected side effects.

I think what I'd expect from SDL is:
* If the first visible window is resizable and not fullscreen, the "start maximized" size is applied (with the correct size).
* Otherwise it's ignored.

If SDL3 still has behaviour similar to #1714, maybe a bug report can be filed there? And any code changes to fix or work around the issue could be applied to SDL's codebase rather than love's.

learn-more: The main window in SDL3 does not seem affected, the compatibility messagebox does.

I have filed an issue: https://github.com/libsdl-org/SDL/issues/9434

slime73: Thanks for investigating!

learn-more: SDL has added the [3.2.0](https://github.com/libsdl-org/SDL/milestone/10) milestone on that ticket.

PR #2047, opened 2024-03-30
Fix building love on Windows for SDL3-3.1.0

learn-more:
Both opengl and vulkan seem to start / run fine with my simple test application on Windows.

### DIFF
--- a/src/love.cpp
+++ b/src/love.cpp
@@ -26,6 +26,10 @@
 
 #ifdef LOVE_BUILD_EXE
 
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+#include <SDL_main.h>
+#endif
+
 // Lua
 extern "C" {
 	#include <lua.h>
--- a/src/modules/window/sdl/Window.cpp
+++ b/src/modules/window/sdl/Window.cpp
@@ -1472,15 +1472,27 @@ void Window::swapBuffers()
 					dwmRefreshRate = (double)info.rateRefresh.uiNumerator / (double)info.rateRefresh.uiDenominator;
 
 				SDL_DisplayMode dmode = {};
+#if SDL_VERSION_ATLEAST(3, 0, 0)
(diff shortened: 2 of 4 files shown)


src/modules/window/sdl/Window.cpp
```diff
@@ -1443,6 +1443,33 @@ bool Window::isMinimized() const
 	return window != nullptr && (SDL_GetWindowFlags(window) & SDL_WINDOW_MINIMIZED);
 }
 
+static void get_displaymode(SDL_DisplayMode* mode, SDL_Window* window)
```
slime73, line 1446: Usually I prefer to keep the #if branches in code where it's used instead of splitting it into its own function, since then we can just delete the other branch once we move to SDL3-only, without having to refactor anything else.

(This also applies to the swapinterval function.)

slime73 (changes requested):

src/modules/window/sdl/Window.cpp
learn-more: And how about writing the code for sdl3, and add a backwards compat function with this same prototype for sdl2?
Then only the function has to be removed when dropping sdl2 support

slime73: I don't think this code really benefits from a function as-is, since you'd have to navigate away from where it's used to see what it does but it doesn't do something that's especially useful outside of that context.

src/love.cpp
```diff
@@ -26,6 +26,10 @@
 
 #ifdef LOVE_BUILD_EXE
 
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+#include "SDL_main.h"
```
slime73, line 30: Should this use system-relative includes (`<SDL_main.h>` instead of `"SDL_main.h"`)?

learn-more: sure

slime73:
Merged, thanks!

PR #2051, opened 2024-04-03
Group liblove and lovedep projects in Visual Studio

learn-more:
This is the end result:

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ project(love)
 set(CMAKE_MODULE_PATH "${love_SOURCE_DIR}/extra/cmake" ${CMAKE_MODULE_PATH})
 set(CMAKE_POSITION_INDEPENDENT_CODE TRUE) # Needed for shared libs on Linux. (-fPIC).
 set(CMAKE_CXX_STANDARD 17)
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)	# Allow grouping projects in Visual Studio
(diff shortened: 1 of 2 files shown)


slime73:
From the linux build failure:
```
CMake Error at extra/cmake/LoveMacros.cmake:11 (get_target_property):
  get_target_property() called with non-existent target "lua51".
Call Stack (most recent call first):
  CMakeLists.txt:1926 (love_group_projects)

-- Configuring incomplete, errors occurred!

CMake Error at extra/cmake/LoveMacros.cmake:13 (if):
  if given arguments:

    "STREQUAL" "INTERFACE_LIBRARY"

  Unknown arguments specified
```

learn-more:
@slime73 This should now be fixed

slime73:
This seems ok to me, but I'd like someone with more cmake experience to look at it as well.

bartbes:
Would it make sense to also separate out the 3rd-party (love_3p) targets?

learn-more:
> Would it make sense to also separate out the 3rd-party (love_3p) targets?

If you prefer that I can group those in a sub-folder as well.
Any suggestions for the name?
I'd think something like `lovedep/third-party` or so?

slime73:
Maybe `liblove/libraries/` makes sense, given that they're in-source dependencies.

learn-more:
image

slime73:
Looks good, thanks!

PR #2052, opened 2024-04-05
Add love.system.getSDLVersion()

learn-more:
This is in preparation of SDL2/SDL3 fun.

### DIFF
--- a/src/modules/system/wrap_System.cpp
+++ b/src/modules/system/wrap_System.cpp
@@ -21,6 +21,7 @@
 // LOVE
 #include "wrap_System.h"
 #include "sdl/System.h"
+#include <SDL_version.h>
 
 namespace love
 {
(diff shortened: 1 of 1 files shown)


slime73:
What use case do you have in mind for this? We aren't planning to ship love 12 with support for both SDL2 and SDL3, we just temporarily allow compiling for either while SDL3 and love 12 are in development so we don't have to do a bunch of work right when SDL3 is released.

learn-more:
I was under the impression there would be sdl2/sdl3 versions for a while, so I figured it might be interesting for developers to have this info available.

I used it to demonstrate running from the same codebase with only changing one cmake var:
image

slime73:
We don't intend to have a split in a released version, the split now is just for C++ development purposes.

PR #2053, opened 2024-04-09
vulkan: fix incorrect scissor when rendering to a canvas

nikeinikei:


src/modules/graphics/vulkan/Graphics.cpp
```diff
@@ -1033,11 +1003,22 @@ void Graphics::setScissor()
...
+	{
+		VkRect2D scissor{};
+		scissor.offset = { 0, 0 };
+		if (renderPassState.isWindow)
+			scissor.extent = swapChainExtent;
+		else
+		{
+			double dpiScale = getCurrentDPIScale();
+
+			scissor.extent.width = static_cast<double>(renderPassState.width) * dpiScale;
```
slime73, line 1016: I think renderPassState.width/height are already in pixels, so they shouldn't need the dpiScale multiplier.

slime73:
Merged, thanks!

PR #2057, opened 2024-04-17
Add support for SDL3 when building with megasource

learn-more:
The lovedep target is renamed from SDL2 to SDL, but the variables used are kept as-is (with SDL3 specific stuff added).
This is so that when the move to SDL3 should be made, SDL2 stuff can be removed.

The linux path is not touched, because I do not know how this should be done exactly.

This PR accompanies love2d/megasource#17

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,7 +98,7 @@ endif()
 
 message(STATUS "Target platform: ${LOVE_TARGET_PLATFORM}")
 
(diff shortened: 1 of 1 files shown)


learn-more:
I am testing out the github builder in my own fork, feedback is welcome:
https://github.com/learn-more/love/pull/1

(The bug that tripped the builds up is sent as PR love2d/megasource#18, non-megasource builders are disabled during testing)

Issue #2068, opened 2024-05-10
Undefined symbols linking for macOS building with CMake


ryandesign:
After working around #2066, building the main branch for macOS using CMake…

```
cmake -S. -Bbuildmac -DCMAKE_FIND_FRAMEWORK=LAST
cmake --build buildmac -j4
```

…fails at link time:

```
Undefined symbols for architecture x86_64:
  "love::apple::getUserDirectory(love::apple::UserDirectory)", referenced from:
      love::filesystem::physfs::Filesystem::getFullCommonPath(love::filesystem::Filesystem::CommonPath) in liblove_filesystem_physfs.a(Filesystem.cpp.o)
  "love::apple::getExecutablePath()", referenced from:
      love::filesystem::Filesystem::getExecutablePath() const in liblove_filesystem_root.a(Filesystem.cpp.o)
  "love::graphics::metal::createInstance()", referenced from:
      love::graphics::Graphics::createInstance() in liblove_graphics_root.a(Graphics.cpp.o)
ld: symbol(s) not found for architecture x86_64
```

I'll submit a PR for this shortly.

Issue #2066, opened 2024-05-09
error: call to member function 'WriteNumber' is ambiguous


ryandesign:
I have seen the warning that "CMake is not an officially supported build system for love on Apple platforms" but I'm trying it anyway.

I got these errors when building the main branch with Apple clang version 14.0.0 (clang-1400.0.29.202):

```
src/libraries/vma/vk_mem_alloc.h:644
6:10: error: call to member function 'WriteNumber' is ambiguous
    json.WriteNumber(allocationCount);
    ~~~~~^~~~~~~~~~~
src/libraries/vma/vk_mem_alloc.h:568
6:21: note: candidate function
void VmaJsonWriter::WriteNumber(uint32_t n)
                    ^
src/libraries/vma/vk_mem_alloc.h:569
3:21: note: candidate function
void VmaJsonWriter::WriteNumber(uint64_t n)
                    ^
```

```
src/libraries/vma/vk_mem_alloc.h:159
70:34: error: call to member function 'ContinueString' is ambiguous
                            json.ContinueString(index++);
                            ~~~~~^~~~~~~~~~~~~~
src/libraries/vma/vk_mem_alloc.h:5657:21: note: candidate function
void VmaJsonWriter::ContinueString(uint32_t n)
                    ^
src/libraries/vma/vk_mem_alloc.h:5663:21: note: candidate function
void VmaJsonWriter::ContinueString(uint64_t n)
                    ^
```

nikeinikei:
It's not directly the problem, but the Vulkan backend shouldn't be included in the macos build anyway (which is the sole user of VMA), you can try excluding it by modifying cmakelists.txt, the dependency on VMA should be gone then as well.

ryandesign:
#2069 works around the problem on macOS by not building the Vulkan parts and instead building the Metal parts, but you would presumably still run into this error if you use a similar compiler on Linux or Windows.

slime73:
> you would presumably still run into this error if you use a similar compiler on Linux or Windows.

I don't think I've heard of anyone having compile problems when using clang on Linux with love 12. Is it definitely an issue?

PR #2069, opened 2024-05-10
Fix CMake build on macOS

ryandesign:
This fixes the CMake build on macOS by fixing #2068 and working around #2066.


slime73:
Merged, thanks!

PR #2078, opened 2024-06-17
Make alext.h optional

DoomHammer:
macOS doesn't seem to have these files but love compiles fine if we just don't include them. Otherwise it fails as the header cannot be found.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,6 +198,8 @@ Please see https://github.com/love2d/megasource
 	find_package(OpenAL REQUIRED)
 	target_include_directories(lovedep::OpenAL INTERFACE ${OPENAL_INCLUDE_DIR})
 	target_link_libraries(lovedep::OpenAL INTERFACE ${OPENAL_LIBRARY})
+  include(CheckIncludeFileCXX)
+  check_include_file_cxx("alext.h" HAVE_OPENAL_ALEXT_H)
 
 	find_package(ModPlug REQUIRED)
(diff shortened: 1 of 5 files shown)


slime73:
We use OpenAL-Soft on macOS which has those files, it sounds like you don't have the right dependencies.

Is there a particular reason for using CMake on macOS instead of the xcode project (which can be used from the command line) or the prebuilt app? The xcode project is the only officially supported build system for macOS, for love.

DoomHammer:
There are two reasons:

1. I wanted to package love for Nix and CMake seems like a nice way to do it
2. I despise XCode and can't really get my way around it

I'll try working with OpenAL-Soft and see if I have better results this way.

BTW: why is the CMake (yet?) unsupported? Do you plan to replace XCode fully at some point or are they meant to be used in parallel?

slime73:
> BTW: why is the CMake (yet?) unsupported? Do you plan to replace XCode fully at some point or are they meant to be used in parallel?

Using the platform's native build system means we don't have to wait for months or years for a meta-build system to support any changes, which is especially important on Apple platforms because they make changes every year. CMake also has very poor support for iOS, you'd want to use Xcode there regardless. Distributing a plain Xcode project also lets people just double click it and build, for both platforms.

> I despise XCode and can't really get my way around it

You can take a look at the repository's github actions file for a couple lines that will build using the xcode project from the command line, and make a distributable .app. But I heavily recommend using the prebuilt .app instead (love 11.5's is uploaded in the Releases section, and love 12 prerelease builds are part of Github Actions artifacts.)

DoomHammer:
I will try that, thank you!

PR #2079, opened 2024-06-17
Install binaries on macOS

DoomHammer:


CMakeLists.txt
```diff
@@ -2138,7 +2138,7 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
 	")
 
 	include(CPack)
-elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+else()
```
slime73, line 2141: It looks like this code is all Linux-specific files, it doesn't seem very relevant to macOS

PR #2088, opened 2024-07-29
Made love.data.newDataView size arg optional

EngineerSmith:
I got annoyed that I had to keep calculating the size lua side when making a new dataview. So, this change makes the 3rd arg [size], optional and is calculated as `love.data.newByteData` calculates the default size when calling `love.data.newByteData(someData)`  [this line](https://github.com/love2d/love/blob/main/src/modules/data/wrap_DataModule.cpp#L88)

I made this change on github directly; and so I haven't tested the change. I'll add a comment once I've tested the build to ensure it works as expected.

Example usage:
```lua
local data = love.data.newByteData(200)
local offset = 100

-- Change:
local dataview = love.data.newDataView(data, offset)
-- Previously:
local dataview = love.data.newDataView(data, offset, data:getSize() - offset)
```

### DIFF
--- a/src/modules/data/wrap_DataModule.cpp
+++ b/src/modules/data/wrap_DataModule.cpp
@@ -57,7 +57,7 @@ int w_newDataView(lua_State *L)
 	Data *data = luax_checkdata(L, 1);
 
 	lua_Integer offset = luaL_checkinteger(L, 2);
-	lua_Integer size = luaL_checkinteger(L, 3);
+	lua_Integer size = luaL_optinteger(L, 3, data->getSize() - offset);
 
 	if (offset < 0 || size < 0)
 		return luaL_error(L, "DataView offset and size must not be negative.");


slime73:
The change seems reasonable to me, but I'm curious what sort of situations you were running into where you wanted to use all the remaining size after an offset

EngineerSmith:
Right now I'm making a chunk of a save file; I precalculated the chunk size/length for all the data I have to insert into the chunk; so I'm treating it like a write to a file which moves the cursor/position along. I've ran into this previously when working with dataviews. So, I thought it would be a "nice to have" than having to store the offset in a different variable.
```lua
-- When quickly prototyping, I don't want to come back and fix this mess. When I can add the source change to handle it for me as I add more to the offset
love.data.newDataView(data, foo + x + y + z, data:getSize() - (foo + x + y + z))
```

I tested it; with this code and looks good.
```lua
local data = love.data.newByteData(200)
local offset = 50

local dv = love.data.newDataView(data, offset)

print(tostring(dv:getSize() == 150), dv:getSize()) -- true   150
```

slime73:
Makes sense, thanks!

Issue #2072, opened 2024-05-30
Physics fixture/shape :setMask bits are inverted in love compared to box2d


snowkittykira:
Currently a fixture/shape needs to have in its mask _every_ category that it doesn't want to collide with, which is the opposite of how it works in box2d

fixture:getMask says "Returns which categories this fixture should NOT collide with. "
https://love2d.org/wiki/Fixture:getMask

the box2d documentation says "The categoryBits flag can be thought of as the fixture saying 'I am a ...', and the maskBits is like saying 'I will collide with a ...'."
https://www.iforce2d.net/b2dtut/collision-filtering

This is awkward for a few reasons:
- an object that only wants to collide with one category needs to specify the names of all the game's other categories in its mask
- it means people reading the box2d documentation will get confused, for example even on the love2d wiki the following diagram repeats the box2d documentation, rather than the love documentation https://love2d.org/wiki/love.physics
- it's rather hard to reason about, because to avoid collision a shape needs to mask every category the other one has, whereas in box2d the logic is simply "if any category matches the mask it collides"

the inversion that love2d does can be seen here and on the corresponding functions in Fixture.cpp:
https://github.com/love2d/love/blob/d5766924ed7fa70ac8f092e725860c51e3fa873d/src/modules/physics/box2d/Shape.cpp#L281

I'm suggesting that in love 12 the API be simplified to do the same thing as the corresponding box2d functionality. If appropriate the functions could be renamed to "setCollisionMask" from "setMask" for backwards compatibility and to make it clearer that the mask says which things an object _will_ collide with.

alex-boulay:
Do we still want the behavior to keep Categories number as target (1 to 16) or their integer values 2^category_number ? 
like could I set by hand setCollisionmask(Shape1, Shape2)  and it would be the same as setCollisionMask(3) ? 
or should setCollisionMask(3) have to set the Category number 3 to true ?

alterae:
i very strongly concur with OP here. flags/masks being inverted is a confusing footgun and makes no sense in light of the behavior of flags/masks in box2d itself

xiaoandtell:
This needs to be fixed to match Box2D specs. It caused me trouble and wasted testing time until I found this open issue. 
To add to the confusion, the code in src/libraries/box2d/dynamics\b2_world_callbacks.cpp and source/modules/physics/box2d/World.cpp matches the Box2D specs.
Can we get an ETA on the fix? Thanks

slime73:
[...]

PR #2090, opened 2024-08-09
Add `setCollisionMask` and `getCollisionMask` to Shape's API

Bruception:
Implements the suggestion in: https://github.com/love2d/love/issues/2072

### DIFF
--- a/src/modules/physics/box2d/Shape.h
+++ b/src/modules/physics/box2d/Shape.h
@@ -161,8 +161,10 @@ public:
 
 	int setCategory(lua_State *L);
 	int setMask(lua_State *L);
+	int setCollisionMask(lua_State *L);
 	int getCategory(lua_State *L);
 	int getMask(lua_State *L);
+	int getCollisionMask(lua_State *L);
 	uint16 getBits(lua_State *L);
 	int pushBits(lua_State *L, uint16 bits);
 
--- a/testing/tests/physics.lua
+++ b/testing/tests/physics.lua
@@ -447,15 +447,15 @@ love.test.physics.Shape = function(test)
   test:assertEquals(nil, yn2, 'check ray 2 y')
 
   -- check filtering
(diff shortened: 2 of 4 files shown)


src/modules/physics/box2d/Shape.h
```diff
@@ -161,8 +161,10 @@ class Shape : public love::physics::Shape
 
 	int setCategory(lua_State *L);
 	int setMask(lua_State *L);
```
Bruception, line 163: Unclear to me if we want to maintain this API, or prefer to deprecate.

testing/tests/physics.lua
```diff
@@ -447,15 +447,15 @@ love.test.physics.Shape = function(test)
   test:assertEquals(nil, yn2, 'check ray 2 y')
 
   -- check filtering
-  test:assertEquals(nil, shape2:getCollisionMask(), 'check no mask')
-  shape2:setCollisionMask(1, 2, 3)
-  test:assertEquals(3, #{shape2:getCollisionMask()}, 'check set mask')
+  test:assertEquals(nil, shape2:getMask(), 'check no mask')
+  shape2:setMask(1, 2, 3)
+  test:assertEquals(3, #{shape2:getMask()}, 'check set mask')
```
Bruception, line 452: Reverting after reading:
https://github.com/love2d/love/tree/main/testing#l%C3%B6vetest

Bruception:
@slime73, is there anything else I should do to get this across the line?

slime73:
I'm wondering about usability - if someone wants an object to collide with everything or nearly everything, they'd need to specify around 16 parameters with the new API whereas with the old one it'd be much less.

alex-boulay:
Previous behavior still works if I'm correct so you can still use it. The real question to me is will we keep the two implementations ? I would prefer to have functions: 
  - clear()/fill()/add(Fixtures...)/remove(Fixtures...)-Mask. 
It would allow better handling especially when dynamically removing collisions for example in cases of events.

PR #2126, opened 2024-12-16
Improve `"justify"` text alignment

steveRoll-git:
The `"justify"` align mode now looks more like proper justified text - the right edges of lines are flush with the given wrap width.

Before:

https://github.com/user-attachments/assets/1e53dca5-9276-4f87-a73d-10029ca307c4

After:

https://github.com/user-attachments/assets/5fca5614-5df9-4b70-aecd-db5043a41fab

### DIFF
--- a/src/modules/graphics/Font.cpp
+++ b/src/modules/graphics/Font.cpp
@@ -508,7 +508,7 @@ std::vector<Font::DrawCommand> Font::generateVerticesFormatted(const love::font:
 				auto end = start + range.getSize();
 				float numspaces = std::count(start, end, ' ');
 				if (width < wrap && numspaces >= 1)
-					extraspacing = floorf((wrap - width) / numspaces);
+					extraspacing = (wrap - width) / (numspaces - 1);
 				else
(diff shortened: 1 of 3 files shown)


src/modules/graphics/Font.cpp
```diff
@@ -507,8 +507,12 @@ std::vector<Font::DrawCommand> Font::generateVerticesFormatted(const love::font:
 				auto start = text.cps.begin() + range.getOffset();
 				auto end = start + range.getSize();
 				float numspaces = std::count(start, end, ' ');
+
+				if (text.cps[range.last] == ' ')
```
slime73, line 511: Should this loop backwards until the first non-space character? In case there are multiple spaces in a row

steveRoll-git: I'm not sure if dealing with trailing spaces is worth it in this case, since text is still affected when there's additional spaces in the beginning, or in the middle

slime73:
Looks good, thanks!

PR #2131, opened 2024-12-31
added hidehomeindicator for iOS

razziefox:
Currently there is no way to control the way the home indicator behaves on iOS without needing to modify the Love source code. This PR aims to add such of an ability via love.conf under `window.hidehomeindicator` and uses SDL3's `SDL_HINT_IOS_HIDE_HOME_INDICATOR` hint. Feel free to let me know any additional changes you'd like to see for this PR.

Example of `window.hidehomeindicator` set to `"2"`
Screen Recording 2024-12-31 at 1 09 27 AM-SD 480p

```lua
function love.conf(t)
    
    t.window.title = "Prismical [dev build 7-18-24]"

    t.window.minwidth = 256
    t.window.minheight = 144
    t.window.width = 256*3
    t.window.height = 144*3
    t.window.resizable = true
    t.window.highdpi = true
    t.window.hidehomeindicator = "2"
    
end
```

### DIFF
--- a/src/modules/love/boot.lua
+++ b/src/modules/love/boot.lua
@@ -180,6 +180,7 @@ function love.init()
 			resizable = false,
 			centered = true,
 			usedpiscale = true,
+			hidehomeindicator = "0",
 		},
 		graphics = {
 			gammacorrect = false,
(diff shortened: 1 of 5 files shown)


slime73:
If you just want to change between `0` (not hidden) and `2` (requires a swipe to be fully visible), switching between windowed mode and fullscreen should already do that without any extra APIs. The only one currently not possible in love without using the FFI or modifying source code is `1`, which might not be as useful for games I think.

razziefox:
Huh, interesting! Truthfully didn't know that. Though it'd still be nice to have a bit of extra control over the home indicator without relying on if fullscreen is set, especially if it may not be known. At least based off what I've seen from the documentation though, as I could've also missed it entirely without realizing.

PR #2133, opened 2025-01-02
Replace NativeFile to use SDL IO functions.

MikuAuahDark:
This removes any platform-specific code from `NativeFile` and rely on SDL3 IO functions diretly.

One issue I found with this is that SDL3 doesn't expose any function to set the IO buffering mode. This is why this change goes into a PR to discuss how should the buffering function behaves.

### DIFF
--- a/src/modules/filesystem/NativeFile.cpp
+++ b/src/modules/filesystem/NativeFile.cpp
@@ -36,6 +36,9 @@
 #include <unistd.h> // POSIX.
 #endif
 
+// SDL
+#include <SDL3/SDL_iostream.h>
+
 namespace love
 {
 namespace filesystem
--- a/src/modules/filesystem/NativeFile.cpp
+++ b/src/modules/filesystem/NativeFile.cpp
@@ -20,21 +20,9 @@
 
 // LOVE
 #include "NativeFile.h"
-#include "common/utf8.h"
 
-#ifdef LOVE_ANDROID
-#include "common/android.h"
-#endif
-
-// Assume POSIX or Visual Studio.
-#include <sys/types.h>
-#include <sys/stat.h>
-
-#ifdef LOVE_WINDOWS
-#include <wchar.h>
-#else
-#include <unistd.h> // POSIX.
-#endif
+// C
+#include <cstring>
 
 // SDL
 #include <SDL3/SDL_iostream.h>
@@ -48,8 +36,10 @@ NativeFile::NativeFile(const std::string &filename, Mode mode)
 : filename(filename)
 , file(nullptr)
 , mode(MODE_CLOSED)
+, buffer(nullptr)
 , bufferMode(BUFFER_NONE)
 , bufferSize(0)
+, bufferUsed(0)
 {
 	if (!open(mode))
 		throw love::Exception("Could not open file at path %s", filename.c_str());
@@ -59,8 +49,10 @@ NativeFile::NativeFile(const NativeFile &other)
 : filename(other.filename)
 , file(nullptr)
 , mode(MODE_CLOSED)
+, buffer(nullptr)
 , bufferMode(other.bufferMode)
 , bufferSize(other.bufferSize)
+, bufferUsed(0)
 {
 	if (!open(other.mode))
 		throw love::Exception("Could not open file at path %s", filename.c_str());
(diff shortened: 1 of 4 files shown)


src/modules/filesystem/NativeFile.cpp
```diff
@@ -264,32 +193,7 @@ bool NativeFile::setBuffer(BufferMode bufmode, int64 size)
...
-		vbufmode = _IOLBF;
-		break;
-	case File::BUFFER_FULL:
-		vbufmode = _IOFBF;
-		break;
-	}
-
-	if (setvbuf(file, nullptr, vbufmode, (size_t) size) != 0)
-		return false;
+	// FIXME: SDL doesn't have option to set buffering.
```
slime73, line 196: Maybe we should make an issue in their repo about this? I feel like it's pretty useful / important, but at the very least we'd get their opinion.

src/modules/filesystem/NativeFile.h
```diff
@@ -25,6 +25,9 @@
 // C
 #include <cstdio>
 
+// SDL
+#include <SDL3/SDL_iostream.h>
```
slime73, line 29: Can this include be moved to `NativeFile.cpp`? And then this can be a `typedef struct SDL_IOStream SDL_IOStream;` or something - I'd just like to minimize how much of SDL is in public module headers when possible.

src/modules/filesystem/NativeFile.cpp
slime73: That being said, for now we could probably implement `BUFFER_LINE` ourselves (like what's done in physfs/File.cpp) and leave the rest as a todo.

src/modules/filesystem/NativeFile.cpp
```diff
@@ -88,49 +88,23 @@ bool NativeFile::open(Mode newmode)
...
 	{
-		if (newmode != MODE_READ)
-		{
-			::close(fd);
-			throw love::Exception("%s is read-only.", filename.c_str());
-		}
+		std::string err = SDL_GetError();
 
-		file = fdopen(fd, "rb");
+		if (err != "")
```
slime73, line 96: Can we remove this `if`? I don't think the error string can be relied on having a specific value.

src/modules/filesystem/NativeFile.cpp
MikuAuahDark: Alright I opened ticket in SDL for this. libsdl-org/SDL#11832

src/modules/filesystem/NativeFile.cpp
MikuAuahDark: I don't even know why I go with `err != ""` instead of `strlen(err) > 0` for this. Will do.

src/modules/filesystem/NativeFile.cpp
MikuAuahDark: Considering that SDL is just low-level wrapper to OS-specific API (such as Windows file `HANDLE` or POSIX file descriptor) and there's no explicit way to set buffering mode on those handle/file descriptors, then the best but unfortunate way to solve this is to implement our own buffering.

MikuAuahDark:
One more thing I forgot to mention is the behavior of `file:getSize()`. It only returns meaningful values (if applicable) once the file has been opened. If the file is closed, it will always return `-1`. This is because it uses [`SDL_GetIOSize`](https://wiki.libsdl.org/SDL3/SDL_GetIOSize) which requires valid `SDL_IOStream*` handle.

slime73:
Our physfs File backend temporarily opens the file to determine its size in that situation: https://github.com/love2d/love/blob/f0a7a19ebcb04927ff16ff420bfa20e1b4fa66a6/src/modules/filesystem/physfs/File.cpp#L178

src/modules/filesystem/NativeFile.cpp
```diff
@@ -311,6 +330,75 @@ File::Mode NativeFile::getMode() const
...
+	{
+		// Entire data fits in the buffer.
+		memcpy(buffer + bufferUsed, data, size);
+		bufferUsed += size;
+		return true;
+	}
+
+	// Could overflow. Write directly.
+	size_t writeSize = (size_t) size;
+	return flush() && (SDL_WriteIO(file, data, writeSize) == writeSize);
```
slime73, line 399: Is a full flush including `SDL_FlushIO` needed here? Or could we make it do most of the work of `flush()` but skip `SDL_FlushIO`?

This will also end up calling `SDL_FlushIO` before `SDL_WriteIO` even if no data is in the buffer, I think.

MikuAuahDark: That one is kinda lazy. I implemented more sophisticated buffered write routines instead.

slime73:
Looks good, thanks!

PR #2147, opened 2025-01-30
Fixed delay in changing window icon on boot

Papaew:
Changing the icon to a custom one (with t.window.icon in conf.lua) results in the icon changing with a delay:
delayed

It is necessary to change the order of window setup in boot.lua: first change the icon, and then open the window:
fixed


slime73:
Looks good, thanks!

PR #2154, opened 2025-02-17
Linux arm64.AppImage

humbertodias:
## Changes
- runner: [ubuntu-22.04, ubuntu-24.04-arm]
- skip Vulkan test on arm64 due to missing dependencies at https://packages.lunarg.com/vulkan/1.3.296/dists/jammy/
- artifact: love-linux-arm64.AppImage
image
https://github.com/humbertodias/love/actions/runs/13359981548

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,12 +80,12 @@ jobs:
         token: ${{ secrets.GITHUB_TOKEN }}
     - name: Zip Test Output (opengl)
       run: |
-        7z a -tzip test-output-linux-opengl.zip love2d-${{ github.sha }}/testing/output/
+        7z a -tzip test-output-linux-opengl-${{ matrix.architecture }}.zip love2d-${{ github.sha }}/testing/output/
     - name: Artifact Test Output (opengl)
       uses: actions/upload-artifact@v4
       with:
         name: test-output-linux-opengl-${{ steps.report1.outputs.conclusion }}
-        path: test-output-linux-opengl.zip
+        path: test-output-linux-opengl-${{ matrix.architecture }}.zip
     # linux opengles tests
     - name: Run Test Suite (opengles)
       env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
       run: sudo apt-get update
     - name: Install Dependencies
       run: |
-        sudo apt-get install --assume-yes build-essential git make cmake autoconf automake \
+        sudo apt install --assume-yes build-essential git make cmake autoconf automake \
                                           libtool pkg-config libasound2-dev libpulse-dev libaudio-dev \
                                           libjack-dev libx11-dev libxext-dev libxrandr-dev libxcursor-dev \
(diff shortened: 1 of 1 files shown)


MikuAuahDark:
Does this actually build for ARM64?

.github/workflows/main.yml
```diff
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 jobs:
   Linux:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-24.04-${{ matrix.architecture }}
```
MikuAuahDark, line 6: IIRC there's only `ubuntu-24.04` for x86-64 and `ubuntu-24.04-arm` for ARM64. This definitely won't work.

MikuAuahDark (changes requested):

.github/workflows/main.yml
humbertodias: now using - runner: [ubuntu-24.04, ubuntu-24.04-arm]

slime73:
Can we make CI build the SDK, similar to the latest comment in that issue report?

humbertodias:
> Does this actually build for ARM64?

Yes, I have successfully tested it now.
love2d-linux-arm64


[love-linux-ARM64.AppImage.zip](https://github.com/user-attachments/files/18818290/love-linux-ARM64.AppImage.zip)

.github/workflows/main.yml
```diff
@@ -100,33 +109,35 @@ jobs:
...
     - name: Artifact Test Output (opengles)
       uses: actions/upload-artifact@v4
       with:
-        name: test-output-linux-opengles-${{ steps.report2.outputs.conclusion }}
-        path: test-output-linux-opengles.zip
+        name: test-output-linux-opengles-${{ runner.arch }}-${{ steps.report2.outputs.conclusion }}
+        path: test-output-linux-opengles-${{ runner.arch }}.zip
     # linux vulkan tests
     - name: Run Test Suite (vulkan)
+      if: runner.arch != 'ARM64'
```
slime73, line 120: Can we make it test vulkan? Debug mode doesn't strictly need to be enabled, although it would be nice.

slime73 (changes requested):

.github/workflows/main.yml
humbertodias: No problem. Enabled

slime73: The vulkan linux arm64 test is failing to start because debug mode is enabled there, which requires the vulkan SDK. So you can either disable the graphics debug environment variable on arm64 linux, or make it build the vulkan SDK.

humbertodias: added for vulkan
`LOVE_GRAPHICS_DEBUG: ${{ runner.arch == 'ARM64' && '0' || '1' }}`

slime73 (approved):

MikuAuahDark (changes requested):
Only some minor changes remaining.

Unfortunately I cannot test if the resulting ARM64 binary run in my Raspberry Pi 5 because it runs Debian 12 (which has older glibc than Ubuntu 24.04) but I'll trust the resulting binary.

MikuAuahDark (approved):

PR #2203, opened 2025-06-27
physfs: Do not throw an exception if the filesystem is already initialized.

mansourmoufid:
On my very old Android phone, when I hit Back and switch to another app, then back to the Love game, it throws this exception. Ignoring this error (PHYSFS_ERR_IS_INITIALIZED) allows me to switch back and forth no problem.


slime73:
PhysFS already being initialized implies other state is still around as well, which means you aren't starting with a clean slate. My first instinct is to say that's not a predictable / reliable / safe situation to be in even if a basic app appears to work.

mansourmoufid:
Yeah I couldn't reproduce it on a newer phone so it might have been the very old version of Android.

MikuAuahDark:
The only reason this happens is if you have LOVE threads lingering around.

mansourmoufid:
> The only reason this happens is if you have LOVE threads lingering around.

😳 Ending all threads fixed it. Thank you!

PR #2211, opened 2025-08-04
Fix crash with setStencilTest for compatibility

itzKiwiSky:
Nothing much, just a way to prevent future crashes with libs that uses stencil.
Just setting to a default value like in 11.x


src/modules/graphics/wrap_Graphics.lua
```diff
@@ -55,6 +55,8 @@ end
 function graphics.stencil(func, action, value, keepvalues)
 	love.markDeprecated(2, "love.graphics.stencil", "function", "replaced", "love.graphics.setStencilMode or setStencilState")
 
+	action = action or "replace" -- make compatible with 11.x by giving default value --
```
slime73, line 58: Can this be `if action == nil then action = "replace" end`? To match the code below it, and to make sure any boolean is still an error like in love 11.

slime73: Also the comment can be removed

slime73 (changes requested):

slime73 (approved):

slime73:
Merged, thanks!

Issue #2198, opened 2025-06-12
Update Box2D to 2.4.2 for LÖVE 12 release


tpimh:
[v2.4.2](https://github.com/erincatto/box2d/releases/tag/v2.4.2) is the final release of version 2 of Box2D, and mostly a bugfix release. I don't suggest switching to a new major version, but such small update should probably be good.

tpimh:
Let's see what's different between Box2D 2.4.1 in LÖVE and upstream 2.4.1:

```diff
--- box2d-2.4.1/box2d/b2_common.h
+++ love2d/box2d/b2_common.h
@@ -34,7 +34,8 @@
 #endif
 
 #define B2_NOT_USED(x) ((void)(x))
-#define b2Assert(A) assert(A)
+void loveAssert(bool test, const char* teststr);
+#define b2Assert(A) loveAssert((A), #A)
 
 #define	b2_maxFloat		FLT_MAX
 #define	b2_epsilon		FLT_EPSILON

--- box2d-2.4.1/box2d/b2_math.h
+++ love2d/box2d/b2_math.h
@@ -24,6 +24,7 @@
 #define B2_MATH_H
 
 #include <math.h>
+#include <cmath>
 
 #include "b2_api.h"
 #include "b2_settings.h"
@@ -31,7 +32,7 @@
 /// This function is used to ensure that a floating point number is not a NaN or infinity.
 inline bool b2IsValid(float x)
 {
-	return isfinite(x);
+	return std::isfinite(x);
 }
 
 #define	b2Sqrt(x)	sqrtf(x)

--- box2d-2.4.1/box2d/Box2D.h
+++ love2d/box2d/Box2D.h
@@ -43,6 +43,7 @@
 #include "b2_time_step.h"
 #include "b2_world.h"
 #include "b2_world_callbacks.h"
+#include "b2_distance.h"
 
 #include "b2_distance_joint.h"
 #include "b2_friction_joint.h"

--- box2d-2.4.1/box2d/common/b2_settings.cpp
+++ love2d/box2d/common/b2_settings.cpp
@@ -27,6 +27,8 @@
 #include <stdarg.h>
 #include <stdlib.h>
 
+#include "common/Exception.h"
+
 b2Version b2_version = {2, 4, 0};
 
 // Memory allocators. Modify these to use your own allocator.
@@ -46,6 +48,12 @@
 	vprintf(string, args);
 }
 
+void loveAssert(bool test, const char* teststr)
+{
+	if (!test)
[...]

PR #2217, opened 2025-09-03
Update Box2D to 2.4.2

tpimh:
Fixes issues #2198 and #2360

### DIFF
--- a/src/libraries/box2d/b2_body.h
+++ b/src/libraries/box2d/b2_body.h
@@ -243,7 +243,7 @@ public:
 
 	/// Get the mass data of the body.
 	/// @return a struct containing the mass, inertia and center of the body.
-	void GetMassData(b2MassData* data) const;
+	b2MassData GetMassData() const;
 
 	/// Set the mass properties to override the mass properties of the fixtures.
 	/// Note that this changes the center of mass position.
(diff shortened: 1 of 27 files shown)


tpimh:
`Joint:getType()` was returning `distance` for [RopeJoint](https://love2d.org/wiki/RopeJoint) since Box2D 2.4.1 (because this is what it is now: https://github.com/love2d/love/commit/fc2cf602f67cbf97286f6b307baff104228e3629).

I removed `rope` value completely, because it became dead code. Maybe it's possible to add some detection logic and return the correct type here:

https://github.com/love2d/love/blob/5903ad8901a1cb3a2bf1d65ab58f3347a5821323/src/modules/physics/box2d/Joint.cpp#L67-L68

tpimh:
Or maybe it's better to just override `getType()` like this https://github.com/tpimh/love/commit/9fa0007a71af2883380b9978ab66d73406753db6?

hexatorum:
the ci for windows can't find the nsis compiler and fails
but other than that error this looks fine to me, can we merge this now

tpimh:
@hexatorum in current state, this PR removes "rope" `JointType` completely (it was broken since the last Box2D update). My proposed solution to restore this functionality is here: https://github.com/tpimh/love/commit/9fa0007a71af2883380b9978ab66d73406753db6. If it looks good to you, I will replace the last commit in this PR with the one from that branch.

tpimh:
I have done some testing, and it seems to work fine.

PR #2219, opened 2025-09-05
Remove deprecation from PHYSFS_getUserDir function

tpimh:
`PHYSFS_getUserDir()` was deprecated for a long time, and given a deprecation warning:

```
love/src/modules/filesystem/physfs/Filesystem.cpp: In member function ‘virtual std::string love::filesystem::physfs::Filesystem::getFullCommonPath(love::filesystem::Filesystem::CommonPath)’:
love/src/modules/filesystem/physfs/Filesystem.cpp:720:62: warning: ‘const char* PHYSFS_getUserDir()’ is deprecated [-Wdeprecated-declarations]
  720 |                 fullPaths[path] = normalize(PHYSFS_getUserDir());
      |                                             ~~~~~~~~~~~~~~~~~^~
In file included from love/src/modules/filesystem/physfs/File.h:29,
                 from love/src/modules/filesystem/physfs/Filesystem.cpp:29:
love/src/libraries/physfs/physfs.h:820:25: note: declared here
  820 | PHYSFS_DECL const char *PHYSFS_getUserDir(void) PHYSFS_DEPRECATED;
      |                         ^~~~~~~~~~~~~~~~~
```

The warning about the deprecated function comes up only on Linux and Android (reported: https://github.com/love2d/love-android/issues/113 and https://github.com/love2d/love-android/issues/146). It's unlikely that this function is going to be removed in physfs 3.x, and even if it is removed from the next release, LÖVE is using modified source of the library and can keep a copy of this function.

I think it's safe to just disable the deprecation warning.

### DIFF
--- a/src/libraries/physfs/physfs.h
+++ b/src/libraries/physfs/physfs.h
@@ -803,8 +803,6 @@ PHYSFS_DECL const char *PHYSFS_getBaseDir(void);
  * \fn const char *PHYSFS_getUserDir(void)
  * \brief Get the path where user's home directory resides.
  *
- * \deprecated As of PhysicsFS 2.1, you probably want PHYSFS_getPrefDir().
- *
  * Helper function.
  *
  * Get the "user dir". This is meant to be a suggestion of where a specific
@@ -819,7 +817,7 @@ PHYSFS_DECL const char *PHYSFS_getBaseDir(void);
  * \sa PHYSFS_getBaseDir
  * \sa PHYSFS_getPrefDir
  */
-PHYSFS_DECL const char *PHYSFS_getUserDir(void) PHYSFS_DEPRECATED;
+PHYSFS_DECL const char *PHYSFS_getUserDir(void);
 
 
 /**


tpimh:
I also noticed that this commit https://github.com/icculus/physfs/commit/bfa7997c671957eb0a340ff1cf3c634e6269904a was not applied when updating PhysFS to 3.2.0. It's a very minor change only related to Solaris, which is not an officially supported platform for LÖVE (and I honestly doubt anyone would ever build this code for Solaris), but I can also include it as a part of this PR for the sake of completeness.

slime73:
I'd rather not modify third party libraries unless it's absolutely necessary. Were you running into a specific problem that this solves?

tpimh:
PhysFS in LÖVE is already a modified version if we compare it to 3.2.0 release:

1. Workaround for Windows 7 runtime compatibility: https://github.com/love2d/love/commit/444ab6311b5e5c19bd65de4f29b876fb9d3124a4
2. Fix compile error when targetting Windows ARM64: https://github.com/love2d/love/commit/cfda1f54258833e02c45cfaa6b331e5f55a5fed8
3. mountRW and friends: https://github.com/icculus/physfs/compare/main...slime73:mountRW

I guess another small change won't hurt. The specific problem that this solves is suppressing the deprecation warning. It can also be supressed with #pragma without modifying the third-party library code, but since this code is already modified in this project, what I did seems to be a much cleaner solution.

slime73:
If we want to get rid of the warning, I think I'd rather remove calls to the function (perhaps by replacing them with OS-specific calls instead).

tpimh:
`__PHYSFS_platformCalcUserDir()` from [physfs_platform_posix.c](https://github.com/icculus/physfs/blob/9d18d36b5a5207b72f473f05e1b2834e347d8144/src/physfs_platform_posix.c#L86) and [physfs_platform_windows.c](https://github.com/icculus/physfs/blob/9d18d36b5a5207b72f473f05e1b2834e347d8144/src/physfs_platform_windows.c#L548) seems solid, and can be just copied, but then again, it would be just duplicated code which in my opinion is worse than a harmless warning.

PR #2220, opened 2025-09-15
Add love.system.getMemorySize

lucenthropic:
Wraps SDL's [`GetSystemRAM`](https://wiki.libsdl.org/SDL3/SDL_GetSystemRAM) which is available since SDL 3.2.0 according to the docs.


slime73:
What's your use-case for this?

lucenthropic:
If our game triggers the error handler, we show a button the player can click to open and pre-fill a bug report as a Google Form with the traceback and some anonymous system/performance info from their machine. It's so we can get a rough idea of the player's machine as extra context for the bug report without requiring the player to manually provide it to us.

slime73:
Merged, thanks!

PR #2226, opened 2025-09-27
Expose more battery and power connection state

### LINKED ISSUE #2223: [REQUEST] Expose controller battery/power level via love.joystick (SDL_JoystickCurrentPowerLevel)
Requesting an API to query a gamepad/joystick’s power level (Unknown/Empty/Low/Medium/Full/Wired). 

Useful for showing “controller low battery” UI and for diagnostics in games that support long sessions and multiple local players.

### Motivation / Use cases
Alert player when a wireless pad is about to die.
Prefer wired pads in tournaments or when recording replays.
Telemetry/debug overlays (e.g., per-player status in local multiplayer).
Parity with SDL, which LÖVE already uses.
Proposed API (Lua)

local level = joystick:getPowerLevel()  -- "unknown" | "empty" | "low" | "medium" | "full" | "wired"

-- Optional convenience:
local seconds = joystick:getEstimatedBatterySeconds() -- if available; else nil

### Behavior / Details
Values mirror SDL’s SDL_JoystickPowerLevel: Unknown, Empty, Low, Medium, Full, Wired.
For devices without a battery or with OS restrictions, return "unknown" (or enum Unknown).
Should not allocate or block; callable every frame.

### SDL reference
SDL_JoystickCurrentPowerLevel(SDL_Joystick*) returns SDL_JoystickPowerLevel (Unknown/Empty/Low/Medium/Full/Wired).

musshorn:
This pull implements the api to call the following:
https://wiki.libsdl.org/SDL3/SDL_GetJoystickPowerInfo
https://wiki.libsdl.org/SDL3/SDL_GetJoystickConnectionState
https://wiki.libsdl.org/SDL3/SDL_GetGamepadPowerInfo
https://wiki.libsdl.org/SDL3/SDL_GetGamepadConnectionState

It's worth noting though, the battery info returned from PowerInfo they have on the wiki:
_You should never take a battery status as absolute truth. Batteries (especially failing batteries) are delicate hardware, and the values reported here are best estimates based on what that hardware reports. It's not uncommon for older batteries to lose stored power much faster than it reports, or completely drain when reporting it has 20 percent left, etc._

I found this in testing, I tried a wireless steam controller on battery and it did not report any battery percent. Other controllers may though. Even when using the SDL testcontroller it did not report anything.

Testing wired though with the following script works as expected
```
function love.draw()
    local sticks = love.joystick:getJoysticks()
    for _, stick in ipairs(sticks) do
        local name = stick:getName()
        local id = stick:getID()
        local battery = stick:getDeviceBatteryPercent()
        local power = stick:getDevicePowerState()
        local connection = stick:getDeviceConnectionState()

        love.graphics.print("name: " .. name, 100, 100)
        love.graphics.print("id: " .. id, 100, 120)
        love.graphics.print("battery power state: " .. power, 100, 140)
        love.graphics.print("battery percent: " .. battery, 100, 160)
        love.graphics.print("connection state: " .. connection, 100, 180)
    end
end

```
image

I suspect this somewhat resolves #2223

### DIFF
--- a/src/modules/joystick/Joystick.h
+++ b/src/modules/joystick/Joystick.h
@@ -144,6 +144,26 @@ public:
 		INPUT_TYPE_MAX_ENUM
 	};
 
+	enum PowerState
+	{
+		POWERSTATE_ERROR = -1,   /**< error determining power status */
+		POWERSTATE_UNKNOWN,      /**< cannot determine power status */
+		POWERSTATE_ON_BATTERY,   /**< Not plugged in, running on the battery */
+		POWERSTATE_NO_BATTERY,   /**< Plugged in, no battery available */
+		POWERSTATE_CHARGING,     /**< Plugged in, charging battery */
+		POWERSTATE_CHARGED,       /**< Plugged in, battery charged */
+		POWERSTATE_MAX_ENUM
+	};
+
+	enum ConnectionState
+	{
+		CONNECTION_INVALID = -1,
+		CONNECTION_UNKNOWN,
+		CONNECTION_WIRED,
+		CONNECTION_WIRELESS,
+		CONNECTION_MAX_ENUM
+	};
+
 	// Represents a gamepad input value, e.g. the "x" button or the left trigger.
 	struct GamepadInput
 	{
@@ -223,6 +243,8 @@ public:
 	virtual bool isSensorEnabled(Sensor::SensorType type) const = 0;
 	virtual void setSensorEnabled(Sensor::SensorType type, bool enabled) = 0;
 	virtual std::vector<float> getSensorData(Sensor::SensorType type) const = 0;
+	virtual PowerState getPowerInfo(int &batteryPercent) const = 0;
+	virtual ConnectionState getConnectionState() const = 0;
 
 	STRINGMAP_CLASS_DECLARE(Hat);
 	STRINGMAP_CLASS_DECLARE(JoystickType);
@@ -230,6 +252,8 @@ public:
 	STRINGMAP_CLASS_DECLARE(GamepadAxis);
 	STRINGMAP_CLASS_DECLARE(GamepadButton);
 	STRINGMAP_CLASS_DECLARE(InputType);
+	STRINGMAP_CLASS_DECLARE(PowerState);
+	STRINGMAP_CLASS_DECLARE(ConnectionState);
 
 	static float clampval(float x);
 
--- a/src/modules/joystick/sdl/Joystick.cpp
+++ b/src/modules/joystick/sdl/Joystick.cpp
@@ -462,6 +462,40 @@ void Joystick::getDeviceInfo(int &vendorID, int &productID, int &productVersion)
 	}
 }
 
+Joystick::PowerState Joystick::getPowerInfo(int& batteryPercent) const
+{
+	// Gets the battery state of a joystick/gamepad
+	Joystick::PowerState powerState;
+	SDL_PowerState batteryState;
+	if (!isGamepad())
+	{
+		batteryState = SDL_GetJoystickPowerInfo(joyhandle, &batteryPercent);
+	}
+	else
+	{
+		batteryState = SDL_GetGamepadPowerInfo(controller, &batteryPercent);
+	}
+	getConstant(batteryState, powerState);
+	return powerState;
+}
+
+Joystick::ConnectionState Joystick::getConnectionState() const
+{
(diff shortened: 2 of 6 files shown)


src/modules/joystick/wrap_Joystick.cpp
```diff
@@ -447,6 +447,43 @@ int w_Joystick_getSensorData(lua_State *L)
 	return (int) data.size();
 }
 
+int w_Joystick_getDeviceBatteryPercent(lua_State *L)
```
slime73, line 450: I think this and `Joystick:getDevicePowerState` should be exposed as one function that returns multiple values (`Joystick:getPowerInfo`), similar to `love.system.getPowerInfo`.

src/modules/joystick/Joystick.h
```diff
@@ -144,6 +144,26 @@ class Joystick : public Object
 		INPUT_TYPE_MAX_ENUM
 	};
 
+	enum PowerState
+	{
+		POWERSTATE_ERROR = -1,   /**< error determining power status */
```
slime73, line 149: I don't think these enums should use `-1` - doing that while putting them in EnumMap/StringMap might not work out well.

Actually an error value shouldn't be part of the enum love uses in the first place I think. I'd rather throw a love exception (which would then be converted to a Lua error) instead, or fall back to `unknown`.

src/modules/joystick/Joystick.h
```diff
@@ -144,6 +144,26 @@ class Joystick : public Object
...
+		POWERSTATE_UNKNOWN,      /**< cannot determine power status */
+		POWERSTATE_ON_BATTERY,   /**< Not plugged in, running on the battery */
+		POWERSTATE_NO_BATTERY,   /**< Plugged in, no battery available */
+		POWERSTATE_CHARGING,     /**< Plugged in, charging battery */
+		POWERSTATE_CHARGED,       /**< Plugged in, battery charged */
+		POWERSTATE_MAX_ENUM
+	};
+
+	enum ConnectionState
```
slime73, line 158: This (and the related methods/functions) could be renamed to `ConnectionType` for consistency with other APIs.

src/modules/joystick/sdl/Joystick.cpp
```diff
@@ -462,6 +462,40 @@ void Joystick::getDeviceInfo(int &vendorID, int &productID, int &productVersion)
 	}
 }
 
+Joystick::PowerState Joystick::getPowerInfo(int& batteryPercent) const
+{
+	// Gets the battery state of a joystick/gamepad
+	Joystick::PowerState powerState;
+	SDL_PowerState batteryState;
+	if (!isGamepad())
```
slime73, line 470: I don't think this needs separate codepaths when the joystick is recognized as a gamepad - using `SDL_GetJoystickPowerInfo` should work in all cases since we always have the joystick handle. SDL has the extra gamepad functions for situations where apps don't use a joystick handle at all.

src/modules/joystick/sdl/Joystick.cpp
```diff
@@ -462,6 +462,40 @@ void Joystick::getDeviceInfo(int &vendorID, int &productID, int &productVersion)
...
+	getConstant(batteryState, powerState);
+	return powerState;
+}
+
+Joystick::ConnectionState Joystick::getConnectionState() const
+{
+	// Gets the connection state of a joystick/gamepad (wired / wireless etc)
+	Joystick::ConnectionState connectionState;
+	SDL_JoystickConnectionState sdlConnectionState;
+	if (!isGamepad())
```
slime73, line 487: Same here with the separate gamepad codepath.

slime73 (changes requested):

src/modules/joystick/wrap_Joystick.cpp
musshorn: Reworked this to better match system:getPowerInfo

src/modules/joystick/sdl/Joystick.cpp
musshorn: No worries, I thought the steam controller I was testing with was reporting as a gamepad when it was on batteries only? Seemed strange but I can't reproduce it now. Simplified code path.

musshorn:
Thanks for the review @slime73 ! Appreciate it. I believe it's all resolved now.

musshorn:
Thanks, those were some obvious ones. I initialize the state now in both functions which makes it a bit simpler.

slime73:
Merged, thanks!

PR #2237, opened 2025-10-19
Buffer: update XYf_STPf_RGBAub DataFormat

TurtleP:
This seems to be a simple typo, but since the texcoords are labeled as `STPf` the `DataFormat` should be `DATAFORMAT_FLOAT_VEC3`. Format `XYf_STPf` reflects this properly.


slime73:
Nice catch, thanks!

PR #2247, opened 2025-11-18
Feat/ios modern vibration api

### LINKED ISSUE #1904: Support `vibrate(seconds)` for iOS

ej-sanmartin:
Fixes #1904
---
**Changes**
- Pass `seconds` duration field from vibrate calls to `void vibrate()` method in `/src/common/ios.mm` that was previously being ignored.
- Implemented duration based `vibrate()` iOS method using iOS 13+ [CHHapticEngine](https://developer.apple.com/documentation/corehaptics/chhapticengine).
---
**Tested**
- Built successfully locally with the following command:
```
xcodebuild -project platform/xcode/love.xcodeproj \
  -scheme love-ios \
  -configuration Debug \
  -sdk iphonesimulator \
  -destination 'platform=iOS Simulator,name=iPhone 17' \
  clean build
```
- Vibration can't be tested through unit tests, as reported by source code https://github.com/love2d/love/blob/main/testing/tests/system.lua#L93-L97

### DIFF
--- a/src/common/ios.mm
+++ b/src/common/ios.mm
@@ -365,57 +365,166 @@ std::string getLoveInResources(bool &fused)
 	return path;
 }
 
+// Static haptic engine with strong reference (persists for app lifetime).
+static CHHapticEngine *sHapticEngine API_AVAILABLE(ios(13.0)) = nil;
+static BOOL sHapticEngineStarted = NO;
+
+// Plays a haptic pattern on the running engine.
+// @param engine  The started haptic engine.
+// @param seconds Duration of the vibration.
+// @return YES on success, NO on failure (caller should use fallback).
+static BOOL playHapticPattern(CHHapticEngine *engine, double seconds)
+	API_AVAILABLE(ios(13.0))
+{
+	NSError *error = nil;
+
+	NSDictionary *hapticDict = @{
+		CHHapticPatternKeyPattern: @[
+			@{
+				CHHapticPatternKeyEvent: @{
+					CHHapticPatternKeyEventType:
+						CHHapticEventTypeHapticContinuous,
+					CHHapticPatternKeyTime: @0.0,
+					CHHapticPatternKeyEventDuration: @(seconds),
+					CHHapticPatternKeyEventParameters: @[
+						// High intensity, correlates to Android's default.
+						@{
+							CHHapticPatternKeyParameterID:
+								CHHapticEventParameterIDHapticIntensity,
+							CHHapticPatternKeyParameterValue: @1.0
+						},
+						// Medium sharpness: 0.0 is rumble, 1.0 is crisp tap.
+						@{
+							CHHapticPatternKeyParameterID:
+								CHHapticEventParameterIDHapticSharpness,
+							CHHapticPatternKeyParameterValue: @0.5
+						}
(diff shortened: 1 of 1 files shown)


src/common/ios.mm
```diff
@@ -364,10 +365,57 @@ - (void)applicationBecameActive:(NSNotification *)note
...
+				};
+
+				CHHapticPattern *pattern = [[CHHapticPattern alloc]
+					initWithDictionary:hapticDict error:nil];
+
+				if (pattern != nil)
+				{
+					id<CHHapticPatternPlayer> player =
+						[engine createPlayerWithPattern:pattern error:nil];
+					[player startAtTime:0 error:nil];
```
slime73: This (and earlier code) should probably have some form of error checking

slime73 (changes requested):

slime73:
> Built successfully locally with the following command:

Can it be tested on a physical device too?

ej-sanmartin:
Thanks for the comment, yes ideally it should be async so it's not blocking the rest of the game's processes. I haven't measured it but most likely there'd be a noticeable performance cost using synchronous implementation that's currently in place.

And I'll test it in my mobile device. With this native API, there's fields that can be adjusted to change how intense and "sharp" the vibration feels. I'll  keep them in them as is since further customization would necessitate us to update Love2D's vibrate API for macOS, which would not be so clean potentially if we want to keep this in parity with the Android version.

ej-sanmartin:
Made the changes and created a minimal Love2D project, which cycles through 3 different vibration calls on iOS.

1st tap tap goes for 1 second, 2nd tap goes for 3 seconds, and the 3rd tap goes for 5 seconds. 4th tap rotates back to the 1 second vibration.

https://github.com/user-attachments/assets/570e88ef-3b40-44e4-af4e-aee43267f9f4

Built successfully with no errors or warnings related to the Haptic Engine initialization (error message did help me identify a missing field though so thanks for suggesting that).

Btw device is an `iPhone 14 Pro Max`, `iOS 26.2`

src/common/ios.mm
ej-sanmartin: SGTM

PR #2253, opened 2025-11-20
Minor Bug Fix: Ensure full traceback is processed by appending a newline

EngineerSmith:
I'm unsure for other users, but on windows `debug.traceback` doesn't return a string with a trailing new line. Which the capture that processes the traceback expects in the default errorhandler. So, the string pattern used, always misses the last line of a traceback from displaying.

So, I've applied a fix by adding a trailing new line if there is one missing. This way the capture pattern now grabs all of the lines of the trace.

### DIFF
--- a/src/modules/love/callbacks.lua
+++ b/src/modules/love/callbacks.lua
@@ -273,7 +273,7 @@ function love.errorhandler(msg)
 	end
 
 	table.insert(err, "\n")
-	
+
 	for l in trace:gmatch("([^\n]+)") do
 		if not l:match("boot.lua") then
 			l = l:gsub("stack traceback:", "Traceback\n")


src/modules/love/callbacks.lua
```diff
@@ -274,6 +274,10 @@ function love.errorhandler(msg)
 
 	table.insert(err, "\n")
 
+	if trace:sub(-1) ~= "\n" then
+		trace = trace .. "\n"
+	end
+	
 	for l in trace:gmatch("(.-)\n") do
```
slime73: Instead of adding a newline, would it also work to change this capture to "everything except newline", like `trace:gmatch("([^\n]+)")`?

EngineerSmith: I've applied your suggestion, after testing it. I didn't want to touch what was already there in fear of breaking some sort of logic; and opted for the safe approach of just applying a newline if one was needed. But, keeping the solution to the pattern is the best fix.

slime73:
Merged, thanks!

PR #2285, opened 2026-02-19
ByteData:setString: Fix Error Handling

TurtleP:
While working on a library of mine, I noticed that the following code did not produce an error:
```lua
local data = love.data.newByteData(4)
data:setString("This is an example")
```
I looked at the implementation of `ByteData:setString` and it always sets the minimum size to either the string length or data size, whichever is smaller. In this example, I'm writing a string with length of 18 into a ByteData with a size of 4. This gets clamped to 4, preventing the error message from showing up. It does make sense to make sure that the size of the data being written is within bounds, but I think the expectation here was to error, not silently fail. I removed the `std::min` call, but if the size is 0, still return nothing back to the Lua stack, since there is nothing to write.

### DIFF
--- a/src/modules/data/wrap_ByteData.cpp
+++ b/src/modules/data/wrap_ByteData.cpp
@@ -51,9 +51,7 @@ int w_ByteData_setString(lua_State *L)
 	const char *str = luaL_checklstring(L, 2, &size);
 	int64 offset = (int64)luaL_optnumber(L, 3, 0);
 
-	size = std::min(size, t->getSize());
-
-	if (size == 0)
+	if (size == 0) 
 		return 0;
 
 	if (offset < 0 || offset + (int64) size > (int64) t->getSize())


pablomayobre:
I feel like this was expected behavior actually, writes the portion of the string that fits into the buffer.

This code in JS (you can run this with Bun or Node.js) could be writte something like:
```js
const buf = Buffer.alloc(4)

buf.write("This is an example")

console.log(buf.toString('utf-8')) //Output: This
```

The output will be "This" with no error message.

Truncation is an expected behavior when trying to write data that doesnt fit on a buffer. You are in charge of validating if there is overflow.

TurtleP:
That's understandable, but what would be the point in having the validation check the bounds, and error, if it's always going to be capped to `Data:getSize()` when it's too big? Either the user validates it or LÖVE does, and given the other `:get*` and `:set*` methods do error in the event of being out-of-bounds, it would make sense that it should error, not truncate.

pablomayobre:
Oh I now see what you mean

The issue is with the lines below:
https://github.com/love2d/love/blob/147d39251c2618852c026f8cadf95f0ffd6a746f/src/modules/data/wrap_ByteData.cpp#L59-L60

This is the code that doesnt get executed.

It also seems like the offset is not taken into account into that `std::min` calculation so the truncation is also incorrect.

There is a decision to be made here on should this method error? or should it truncate?

MikuAuahDark:
Since the other `:set*` methods errors if it doesn't fit, then I believe it makes sense to also error in this case.

slime73:
I think we'd need to add an optional size (and perhaps source offset) parameter if we do this, so that people can copy a subset of a string into the ByteData without allocating a new string.

pablomayobre:
So this would be `ByteData:setString(bufferOffset, string, startIndex, length)` or something like that?

PR #2309, opened 2026-04-20
love.window.setPosition: add param waitForSync

shelvacu:
if waitForSync is false, then SDL_SyncWindow is not called, and so the method may return before the requested changes actually happen.

This is backwards-compatible with previous behaviour

- [x] I confirm that any code in this pull request that is not authored by me has proper attribution and licensing included, and that this pull request does not include any output from generative AI / LLM tools.


slime73:
We might look at sync vs async behaviour more closely for love 13, but for now I've merged this change. Thanks!

Issue #1764, opened 2022-01-13
[SUGGESTION] Consider to use MojoAL instead of OpenALSoft.


sharkist:
https://icculus.org/mojoAL/

Pros:
+ is licensed under the zlib license.
+ identical api from OpenALSoft 
+ get rid of OpenAL32.dll
+ written in C, in a single source file (.c), and can be glued to love2d source-code without any new dlls.
+ requires SDL2 only.

Also i hope, change to MojoAL will be possible without any problems or modifications in source-code.

slime73:
I do want to switch to MojoAL instead of Apple's OpenAL implementation for iOS. But on other platforms, OpennAL-Soft exposes some features love uses - for example audio effects (EFX) and filters.

slime73:
It should be feasible to have a compile flag to choose MojoAL instead of OpenAL-Soft / system OpenAL, at least.

slime73:
I tried MojoAL and it blows up when trying to seek a streaming source. So any use of it in love will have to wait until it's more robust.

0azizi0:
There is a library called MINIAUDIO, it seems good and it's actively developed. Not sure if the library fits LOVE or not, take a look at it.

https://miniaud.io

Ghorthalon:
OpenAL-Soft works on iOS doesn't it? Why not use it there as well? Then we'd also get effects like reverb, HRTF, etc. on iOS as well. Would be quite useful.

MikuAuahDark:
OpenAL-soft works on iOS yes. Unfortunately, licensing issue prevents the usage of OpenAL-soft in iOS.

slime73:
In particular, OpenAL-Soft uses the LGPL license which is incompatible with iOS App Store distribution and other related parts of iOS app packaging and installation.

Ghorthalon:
Ah sorry, I missed that bit. That's really unfortunate and quite disappointing... Thanks for the clarification.

ericoporto:
> tried MojoAL and it blows up when trying to seek a streaming source

Can you still reproduce this?

slime73:
I haven't tried recently, but I did file a bug report about it which got solved.

slime73:
MojoAL doesn't currently support SDL3, so love isn't able to use or test it at all at the moment.

MikuAuahDark:
Related issue: icculus/mojoAL#29

PR #2310, opened 2026-04-22
replace OpenAL with mojoAL (SDL3)

### LINKED ISSUE #1234: CLI argument consistency

> game arg1

> love . arg1

choonarine:
- [x] I confirm that any code in this pull request that is not authored by me has proper attribution and licensing included, and that this pull request does not include any output from generative AI / LLM tools.

## Description

[MojoAL](https://github.com/icculus/mojoAL/tree/a32b2b416eeeea5174ebddbeda605307cd029831) was recently updated and is now suitable for SDL3. I've tested some of the audio features and they work well.
I tested it on Windows 11, but unfortunately I don't have any Apple devices and was unable to test it on other iOS or macOS devices, so I'm unsure whether it works well on them.

I know it's a major change and unspecified errors can be caused by a lack of validation, so it may still be unstable.

However, if it is merged into the main branch, it will resolve the issues caused by the original OpenAL library (License issues, etc.), and OpenAL wrapper should be removed from [megasource](https://github.com/love2d/megasource)!

(note: sorry for my poor english ;( It's not my first language)

## Related Issues

#1764

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,7 +104,6 @@ message(STATUS "Target platform: ${LOVE_TARGET_PLATFORM}")
 add_library(lovedep::SDL INTERFACE IMPORTED)
 add_library(lovedep::Freetype INTERFACE IMPORTED)
 add_library(lovedep::Harfbuzz INTERFACE IMPORTED)
-add_library(lovedep::OpenAL INTERFACE IMPORTED)
 add_library(lovedep::Modplug INTERFACE IMPORTED)
 add_library(lovedep::Theora INTERFACE IMPORTED)
 add_library(lovedep::Vorbis INTERFACE IMPORTED)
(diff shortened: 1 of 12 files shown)


choonarine:
🤦🏼🤦🏼🤦🏼🤦🏼
I’ve been mostly working in a separate Git hosting service, so I didn’t realize GitHub had added this AI slop review feature... (and I CANNOT EVEN REMOVE those reviews myself)
If it’s at all possible, I’d really appreciate it if you could remove those reviews. I sincerely apologize for the inconvenience, and thank you very much in advance.

slime73:
Sorry, we don't have copilot enabled at all in the project so we're trying to figure out how to prevent this. It seems like it might be a new github "feature" to use it everywhere even when explicitly disabled...

slime73:
As for the change itself, we shouldn't unconditionally replace OpenAL Soft with MojoAL since many of love's audio APIs (such as effects) won't work with the latter. But adding a way to opt into using MojoAL when compiling love is fine, as long as it doesn't introduce regressions in the APIs it does support.

choonarine:
I’ll update this MojoAL PR to be optional as soon as possible.
I’ll work on improving compatibility with OpenAL APIs that are not fully supported in MojoAL where possible, though it may take some time.
Until then, I’ll leave this PR as a draft.

Thanks for the review! I’ll come back with a improved implementation.

PR #2320, opened 2026-05-17
Fix zlib decompression incomplete data check

### LINKED ISSUE #1234: CLI argument consistency

### LINKED ISSUE #2317: zlib deflate decompression sometimes doesn't work on Linux
This Lua code compresses and decompresses a file. When run with this 390 byte file ([keopatterns.lua.txt](https://github.com/user-attachments/files/27560779/keopatterns.lua.txt)), it fails to be decompressed. Small modifications to the file will still result in the same error, but may also succeed seemingly at random.

```lua
function love.run()
    local file, out, compressed

    out = io.open("out.bin", "w+")
    file = io.open("keopatterns.lua.txt", "r")
    compressed = love.data.compress("string", "deflate", file:read("*a"), 9)
    print(pcall(love.data.decompress, "string", "deflate", compressed))

    out:write(compressed)
    out:close()

    return function()
        return 0
    end
end
```
Output: `false   Could not decompress zlib/gzip-compressed data (error code: -3).`
(The code also writes the raw compressed bytes to out.bin)

[On this line](https://github.com/love2d/love/blob/7bdef86d3c0eca8edcf1c85d796238bd5720f5ba/src/modules/data/Compressor.cpp#L244), inflate returns Z_BUF_ERROR (-5) and stream.avail_in == 0, so Z_DATA_ERROR (-3) is returned.

I was not able to replicate this with my own C program that calls the zlib functions.

This only seems to be broken on Linux (Fedora 44). It works correctly in Windows (but zlib seems to compress the file differently on Windows). I have not tried other platforms.

Syyrion:

## Description

Fixes a bug with zlib deflate decompression functions falsely detecting compressed data as incomplete.

## Related Issues

Fixes #2317

### DIFF
--- a/src/modules/data/Compressor.cpp
+++ b/src/modules/data/Compressor.cpp
@@ -246,7 +246,7 @@ private:
 		if (err != Z_STREAM_END)
 		{
 			inflateEnd(&stream);
-			if (err == Z_NEED_DICT || (err == Z_BUF_ERROR && stream.avail_in == 0))
+			if (err == Z_NEED_DICT || (err == Z_BUF_ERROR && stream.avail_out > 0 && stream.avail_in == 0))
 				return Z_DATA_ERROR;
 			return err;
 		}


slime73:
Merged, thanks!

PR #2330, opened 2026-06-24
Add 'inverseOf' to love.Transform

erinmaus:

## Description
Add 'inverseOf' that copies the inverse of the parameter (`source`) to the transform `self`. This is to provide a means to prevent generating garbage when getting the inverse of a transform. Generally you can use all of the `Transform` methods without generating garbage as it stands... except when computing the inverse. See Discord discussion: https://discord.com/channels/329400828920070144/516361887840075812/1517506663379238993

I'm unsure of the name - took Sasha's suggestion from Discord.

Personally, my use case is for a three/four dimensional `inverseTransformPoint` sort of equivalent (e.g., `project` and `unproject` equivalents for 3D) using `love.Transform` without generating garbage. I'd like to use the underlying `love.Transform` when `send`ing to shaders, `setProjection`/`applyTransform`, etc and not have to marshal between a Lua/LuaJIT-native `mat4`-equivalent type and `love.Transform`. Can imagine lots of other uses - skeletal animation (e.g., a Lua/LOVE Spine runtime) etc. And a way to avoid generating garbage is always cool. :)

If this is too niche, that's fine. Just thought I'd propose it!

## Related Issues
N/A

### DIFF
--- a/src/modules/math/wrap_Transform.cpp
+++ b/src/modules/math/wrap_Transform.cpp
@@ -145,9 +145,7 @@ int w_Transform_inverseOf(lua_State *L)
 {
 	Transform *destination = luax_checktransform(L, 1);
 	Transform *source = luax_checktransform(L, 2);
-	Transform *inverseSource = source->inverse();
-	destination->setMatrix(inverseSource->getMatrix());
-	inverseSource->release();
+	destination->setMatrix(source->getMatrix().inverse());
 	lua_pushvalue(L, 1);
 	return 1;
 }
--- a/src/modules/math/wrap_Transform.cpp
+++ b/src/modules/math/wrap_Transform.cpp
@@ -141,6 +141,17 @@ int w_Transform_inverse(lua_State *L)
 	return 1;
 }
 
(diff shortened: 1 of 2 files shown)


src/modules/math/wrap_Transform.cpp
```diff
@@ -141,6 +141,17 @@ int w_Transform_inverse(lua_State *L)
 	return 1;
 }
 
+int w_Transform_inverseOf(lua_State *L)
+{
+	Transform *destination = luax_checktransform(L, 1);
+	Transform *source = luax_checktransform(L, 2);
+	Transform *inverseSource = source->inverse();
+	destination->setMatrix(inverseSource->getMatrix());
```
slime73: I think this can be changed to
```cpp
destination->setMatrix(source->getMatrix().inverse());
```

Which avoids creating a temporary heap-allocated Transform object.

erinmaus: Sure thing, updated!

slime73:
Merged, thanks!

Issue #2331, opened 2026-06-24
[Request] Support for QOI texture format in love.graphics.newImage.


sandrarium:
Can you guys implement it? Why?
I need a faster image loader/decoder for my game.

Extremely fast decoding:
- QOI is one of the fastest lossless image formats available, often significantly faster than PNG.

Reasonably small file sizes:
- while not as compressed as PNG in all cases, QOI still provides good lossless compression with far lower CPU cost.

Very small and simple decoder:
- the reference implementation is ~300 lines of C with no external dependencies, making it easy to integrate and maintain.

Great for game assets:
- fast decoding reduces loading times and CPU usage, which is especially beneficial for mobile devices or games that load many textures dynamically.

Potential benefits for Love2D:
- Faster loading times when using love.graphics.newImage
- Lower CPU usage during startup or dynamic asset loading
- A simple, dependency‑free format that fits well with Love2D's lightweight philosophy
- A practical alternative to PNG for developers who want faster load performance

Implementation notes:
- The official QOI implementation is MIT‑licensed
- The decoder is tiny and straightforward to integrate
- github decoder/encoder here: [https://github.com/phoboslab/qoi](https://github.com/phoboslab/qoi)
- webpage: [https://qoiformat.org/](https://qoiformat.org/)

slime73:
In terms of loading speed, native compressed texture formats (e.g. DXT1, DXT5, BC7, ETC2) in file formats like .dds are going to be the fastest by far. Do those suit your needs?

sandrarium:
DXT/BC/ETC are lossy GPU compression formats designed for 3D textures, not pixel‑perfect 2D art.
They introduce block artifacts, color bleeding and smoothing, which breaks the precision required for pixel art.

QOI, on the other hand, is lossless and extremely fast to decode on the CPU, making it a better fit for 2D engines like Love2D where visual fidelity of individual pixels matters.

slime73:
Do you have more specific details about the loading time issues you're facing, and the platforms you're using and the type of game you're making? Also, are you using threaded loading (via an async loading library or manually)?

As for QOI, what common tools support it as an export format?

I'm not particularly against a format that's geared toward fast loading, but understanding more of the context will be useful for making informed decisions.

Sgt-Spaghetti:
[...]

PR #2338, opened 2026-07-15
Feat: QOI image format

### LINKED ISSUE #1234: CLI argument consistency

TotoEnF5:

## Description
Adds a FormatHandler for loading, decoding and encoding images in the [QOI format](https://qoiformat.org/).

## Related Issues
Implements #2331


src/modules/image/magpie/QOIHandler.cpp
```diff
@@ -0,0 +1,118 @@
...
+	uint32_t width;      // image width in pixels (BE)
+	uint32_t height;     // image height in pixels (BE)
+	uint8_t  channels;   // 3 = RGB, 4 = RGBA
+	uint8_t  colorSpace; // 0 = sRGB with linear alpha, 1 = all channels linear
+};
+
+bool QOIHandler::canDecode(Data *data)
+{
+	QOIHeader header;
+	const unsigned char *buffer = (const unsigned char *) data->getData();
```
slime73: canDecode should return false if `data->getSize() < sizeof(QOIHeader)`, otherwise there's a chance of reading an invalid memory address

src/modules/image/magpie/QOIHandler.cpp
```diff
@@ -0,0 +1,118 @@
...
+	memcpy(header.magic, buffer, 4 * sizeof(char));
```
slime73: I think this function can be simplified if the result of `data->getData()` is cast to a `QOIHeader *`, then it doesn't need any memcpy

slime73 (changes requested):

TotoEnF5:
Both changes implemented!

slime73:
Merged, thanks! I'll handle the xcode project files in a followup commit.

PR #2339, opened 2026-07-18
Add WebP Image Support

MikuAuahDark:

## Description
This PR added WebP image support using my [simplewebp](https://github.com/MikuAuahDark/simplewebp) library.

There was an attempt to add WebP image support (see related issue below). Unfortunately, it uses libwebp which are complicated to setup in cross-platform manner and slime prefer a single-header library for it (https://github.com/love2d/love/pull/4#issuecomment-1109930714).

So this one uses simplewebp, a single-header WebP decoder, supporting both lossy and lossless WebP. Output is bit-exact when compared with libwebp.

There's only 1 limitation: Animated WebP is currently unsupported. It just refuse to load them rather than loading the first frame.

## Related Issues
* #4 - First attempt on adding WebP image support.

### DIFF
--- a/src/modules/image/magpie/WEBPHandler.cpp
+++ b/src/modules/image/magpie/WEBPHandler.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2006-2025 LOVE Development Team
+ * Copyright (c) 2006-2026 LOVE Development Team
  *
  * This software is provided 'as-is', without any express or implied
  * warranty.  In no event will the authors be held liable for any damages
--- a/src/modules/image/magpie/WEBPHandler.cpp
+++ b/src/modules/image/magpie/WEBPHandler.cpp
@@ -63,7 +63,7 @@ WEBPHandler::DecodedImage WEBPHandler::decode(Data *data)
 
 	unsigned char *buf = new (std::nothrow) unsigned char[imagesize];
 	if (!buf)
-		throw love::Exception("Could not decode WebP image (%s)", simplewebp_get_error_text(SIMPLEWEBP_ALLOC_ERROR));
+		throw love::Exception("Could not decode WebP image (not enough memory)");
 
 	err = simplewebp_decode(swebp, buf, nullptr);
 	simplewebp_unload(swebp);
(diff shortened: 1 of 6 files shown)


slime73:
How does decoding performance compare with png files in love?

MikuAuahDark:
I have this code:
```lua
function measure(path)
	local COUNT = 100
	local t = love.timer.getTime()
	for i = 1, COUNT do
		local image = love.image.newImageData(path)
	end
	t = love.timer.getTime() - t
	print("avg", path, t * 1000 / COUNT, "ms")
end

function love.load()
	measure("a5.png") -- baseline
	measure("a5.webp") -- lossless
	measure("a!.webp") -- lossy
end
```

The results are (with 2 runs for warmup):
```
avg     a5.png  40.604771       ms (baseline)
avg     a5.webp 67.306112       ms (lossless)
avg     a!.webp 43.146299       ms (lossy)
```

However, I'm converting this to draft back because I just found a bug with the lossless decoding.

MikuAuahDark:
Got the problem fixed upstream so here's new benchmark result. Same code as above with 200 count instead of 100.
```
avg     a5.png  41.4727405      ms (PNG baseline)
avg     a5.webp 69.0992685      ms (WebP lossless)
avg     a!.webp 42.7562545      ms (WebP lossy)
```

Another thing I forgot to mention is the file sizes:
* a5.png is 1.41MB (baseline)
* a5.webp is 1.01MB (lossless)
* a!.webp is 185KB (lossy)

All image has resolution of 2100x900.

Although to be honest simplewebp is meant for easiest integration, not necessarily performant.

src/modules/image/magpie/WEBPHandler.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+namespace image
+{
+namespace magpie
+{
+
+bool WEBPHandler::canDecode(Data *data)
+{
+	simplewebp *swebp = nullptr;
+	simplewebp_error err = simplewebp_load_from_memory(data->getData(), data->getSize(), nullptr, &swebp);
```
slime73, line 44: How much work does this do? Is it just parsing a header, or something more complicated / time-consuming?

MikuAuahDark: That only parses the file header to gather information about the image dimensions, whetever it's lossy or lossless, and (if lossy) whetever alpha channel present.

The actual decoding part is only done in `simplewebp_decode`.

slime73:
Merged, thanks!

PR #2342, opened 2026-07-23
Relax shader storage requirements

erinmaus:

## Description
* Provide an opt-in feature to relax shader storage requirements for users who know what they're doing.
* Refactor compute barriers in OpenGL and Vulkan to support vertex/pixel shaders. Barriers are not needed on Metal 2 (but would be needed on Metal 3 and higher).
* Add barriers around draws **if and only if** a vertex or pixel shader has writable images or buffers.

To support this feature better, also:
* Added `vertexwrite`, `pixelwrite`, ~~and `imageatomics`~~ features to `Graphics::Capabilities` and `love.graphics.isSupported()`

`vertexwrite` and `pixelwrite` indicate a platform can write to both buffers and images in vertex and/or pixel shaders, respectively. ~~`imageatomics` indicates a platform can perform atomics on an `image` in a shader.~~

`vertexwrite` and `pixelwrite` are `true` if:

* On OpenGL and OpenGL ES platforms if GLSL 4 is supported (exposes `glMemoryBarrier`, which is needed around draw calls when using writable buffers)
* On Vulkan platforms, `VkPhysicalDeviceFeatures` structure query returns `true` (`vertexPipelineStoresAndAtomics` for `vertexwrite` and `fragmentStoresAndAtomics` for `pixelwrite`)
* On Metal if running on an Apple 4 device or newer (Apple 3 supports read/write to buffers, while Apple 4 supports read/write to images - so to support the feature, we need Apple 4 devices)

~~`imageatomics` is `true` if:~~

* ~~On OpenGL and OpenGL ES platforms if GLSL 4 is supported. Atomics are supported wherever write is supported.~~
* ~~On Vulkan platforms. Atomics are supported wherever write is supported.~~

~~However, `imageatomics` is false on Metal platforms. This feature is not supported unless using Metal 3.1 or higher. If a shader is compiled for MSL 3.1 on Metal 2, it won't work either (will get an error the functions are missing).~~

Here is an ZIP of a LOVE showing a demo of OIT (click "samples/graphics/oit/test.lua", then select a GLTF), which is enabled using this PR: [oit-example.zip](https://github.com/user-attachments/files/30302797/oit-example.zip)

I tested this on Vulkan/OpenGL on Windows 11 (RTX 3060 Ti laptop) and Metal on macOS 26.5.2 (M3 Max MacBook Pro).

**edit:** Removed references to `imageatomics`.

## Related Issues

See discussion in Discord: https://discord.com/channels/329400828920070144/516361887840075812/1528762120496353363

(Also - I am not attached to names of things used in this PR or any copy feel free to suggest alternatives for clarity, etc.)


src/modules/graphics/metal/Graphics.mm
```diff
@@ -2250,8 +2250,24 @@ static inline void advanceVertexOffsets(const VertexAttributes &attributes, Buff
...
+		capabilities.features[FEATURE_PIXEL_WRITE] = true;
+	}
+	else
+	{
+		capabilities.features[FEATURE_VERTEX_WRITE] = false;
+		capabilities.features[FEATURE_PIXEL_WRITE] = false;
+	}
+
+	// Not supported until Metal 3
+	capabilities.features[FEATURE_IMAGE_ATOMICS] = false;
```
slime73: Since image atomics are format-dependent I think this should be part of `PixelFormatUsageFlags` and set up for each format (on other platforms at least). But that's a large change, it can be done separately from the storage buffers/textures in non-compute shaders change, and I might want to do it myself

erinmaus: I can drop this feature boolean and then folks can use PixelFormatUsageFlags in the future.

erinmaus: Addressed in latest commit (`d8430c3`).

src/modules/graphics/opengl/Graphics.cpp
```diff
@@ -584,6 +594,13 @@ void Graphics::draw(const DrawIndexedCommand &cmd)
 	VertexAttributes attributes;
 	findVertexAttributes(cmd.attributesID, attributes);
 
+	GLbitfield preDrawBarriers = 0;
+	GLbitfield postDrawBarriers = 0;
+
+	shaderBarriers((Shader *)Shader::current, preDrawBarriers, postDrawBarriers);
```
slime73: The return value of `shaderBarriers` is false if there are unbound writable resources, which is a situation that can cause undefined behaviour or crashes on various platforms if the GPU runs the shader - so I think it should early-return here (and in other similar places) before the draw happens if `shaderBarriers` returns false.

Ideally it would cause a user-facing error instead of just skipping the draw, but I'd want to refactor some things to allow for that since the validation for that is more useful if it's run earlier than the actual backend drawing code, so the early return should be fine for now.

Also, the RenderCommandEncoder variant of `applyShaderUniforms` in the Metal backend should be changed to be like the ComputeCommandEncoder variant, where it returns an `allWritableVariablesSet` boolean, which the drawing code would check and then early-return if it's false.

slime73 (changes requested):

src/modules/graphics/opengl/Graphics.cpp
erinmaus: OK, makes sense. I updated all draw paths (Vulkan, OpenGL, and Metal) to return early if any writable buffers are unset.

slime73 (approved):

slime73:
Merged, thanks!

Issue #2336, opened 2026-07-12
[Request] Unclamped max gain


NyakoFox:
The project that I work on sometimes requires me to reimplement behavior from another engine; and in this case, that other engine has unclamped max gain.

OpenALSoft does have support for this, with `AL_SOFT_gain_clamp_ex`. It would be very nice if LÖVE had a way to disable the upper range check in `setVolumeLimits`'s wrapper.

I've considered using FFI to poke around myself, but as far as I'm aware, I can't grab a `ALuint` from a Source.

I know this is behavior not everyone wants, since gain is limited for a reason; which is why I think something opt-in would be nice (like `conf.lua` maybe?)

slime73:
Addressed in https://github.com/love2d/love/commit/bd366341439dbc088c7f92964bd0e13402aefc5a

PR #2345, opened 2026-07-31
unclamped gain #2336

TimeyDingo:
- [yes ] I confirm that all code in this pull request is either authored by me or has proper attribution and licensing included, and that this pull request does not include any output from generative AI / LLM tools.

## Description

- Simply removed a check in w_Source_setVolumeLimits which prevented the user from setting vmax to a number higher then 1
- Tested in custom love2d project on Ubuntu Linux and confirmed that the backend properly bumps the audio above 1 (when the limit is set higher)
- Modified test suite to test for the case where the volume limit is higher then 1

## Related Issues
Feature #2336 
https://github.com/love2d/love/issues/2336

### DIFF
--- a/src/modules/audio/wrap_Source.cpp
+++ b/src/modules/audio/wrap_Source.cpp
@@ -273,7 +273,7 @@ int w_Source_setVolumeLimits(lua_State *L)
 	Source *t = luax_checksource(L, 1);
 	float vmin = (float)luaL_checknumber(L, 2);
 	float vmax = (float)luaL_checknumber(L, 3);
-	if (vmin < .0f || vmin > 1.f || vmax < .0f || vmax > 1.f)
+	if (vmin < .0f || vmin > 1.f || vmax < .0f)
 		return luaL_error(L, "Invalid volume limits: [%f:%f]. Must be in [0:1]", vmin, vmax);
 	t->setMinVolume(vmin);
 	t->setMaxVolume(vmax);
--- a/src/modules/audio/null/Source.cpp
+++ b/src/modules/audio/null/Source.cpp
@@ -176,6 +176,9 @@ float Source::getMinVolume() const
 
 void Source::setMaxVolume(float volume)
 {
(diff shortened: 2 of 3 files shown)


src/modules/audio/wrap_Source.cpp
```diff
@@ -273,7 +273,7 @@ int w_Source_setVolumeLimits(lua_State *L)
 	Source *t = luax_checksource(L, 1);
 	float vmin = (float)luaL_checknumber(L, 2);
 	float vmax = (float)luaL_checknumber(L, 3);
-	if (vmin < .0f || vmin > 1.f || vmax < .0f || vmax > 1.f)
+	if (vmin < .0f || vmin > 1.f || vmax < .0f)
```
slime73, line 276: love uses OpenAL Soft (which always supports the `AL_SOFT_gain_clamp_ex` extension) most of the time, but it doesn't on iOS - and I believe when the extension isn't supported, the `alSourcef` call inside `Source::setMaxVolume` will fail.

I think it'd be better if it clamps it to 1 in that situation. Something like this at the top of `Source::setMaxVolume`'s function:
```cpp
if (!alIsExtensionPresent("AL_SOFT_gain_clamp_ex"))
	volume = std::min(volume, 1.0f);
```

It could also check `#ifdef AL_SOFT_gain_clamp_ex` instead of the `alIsExtensionPresent` call there, I don't have much of a preference and people don't hot-swap the OpenAL dll with a different implementation, so it should be equivalent to the `alIsExtensionPresent` call.

slime73 (changes requested):

src/modules/audio/wrap_Source.cpp
TimeyDingo: I used the second method which was easier to implement of #ifdef's, passed all of the test suite on my local machine

slime73 (approved):

slime73:
Merged, thanks!

PR #2346, opened 2026-08-01
vulkan: closes #2289

### LINKED ISSUE #2289: [12.0] `captureScreenshot` does not work with Vulkan
Testing code:

```lua
local IMAGE = nil
local a = 0

function love.update(dt)
    a = a + dt
end

function love.keypressed(key)
    if key == "f" then
        love.graphics.captureScreenshot(function(image_data)
            IMAGE = love.graphics.newImage(image_data)
            a = 0
        end)
    end
end

function love.draw(dt)
    love.graphics.line(0, 0, 800, 800)

    if IMAGE ~= nil then
        love.graphics.draw(IMAGE, a * 30, a * 10)
    end
end
```

Press `f` to take a screenshot.

Using the Vulkan renderer, the `love.graphics.captureScreenshot` function fails to work, causing strange (inconsistent) results between platforms, whether it's just a transparent image or a part of some texture.

This code works fine on 11.5, and will work properly if you use `--renderers opengl`, but if the renderer is Vulkan, it will not work.

nikeinikei:

## Description
Closes #2289

Moving calling the callbacks untill after the commandBuffer is available again, such that calling love.graphics functions inside the screenshot callback is working as expected.

### DIFF
--- a/src/modules/graphics/vulkan/Graphics.cpp
+++ b/src/modules/graphics/vulkan/Graphics.cpp
@@ -392,7 +392,7 @@ void Graphics::discard(const std::vector<bool> &colorbuffers, bool depthstencil)
 	startRenderPass();
 }
 
-void Graphics::submitGpuCommands(SubmitMode submitMode, void *screenshotCallbackData)
+image::ImageData *Graphics::submitGpuCommands(SubmitMode submitMode)
 {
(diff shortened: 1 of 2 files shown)


slime73:
Merged, thanks!

Issue #1996, opened 2023-12-16
Request: For future iOS release, please set SupportsOpeningDocumentsInPlace to true


SugarRayLua:
Will let users easily add projects to their love2d documents folder from other folders on their iOS device instead of needing to connect to iTunes on their Mac to do the same. 

Thanks!

PR #2349, opened 2026-08-05
SupportOpeningDocsInPlace to true #1996

TimeyDingo:
- [yes ] I confirm that all code in this pull request is either authored by me or has proper attribution and licensing included, and that this pull request does not include any output from generative AI / LLM tools.

## Description

-Changed from false to true to fix issue #1996

https://developer.apple.com/documentation/bundleresources/information-property-list/lssupportsopeningdocumentsinplace


slime73:
Has this been tested? I think the last time I looked into this it seemed like SDL might not be set up to handle that situation properly, but that was back when SDL2 was being used and a lot has changed in its code since then, so I just want to make sure it's working as expected now.

PR #2350, opened 2026-08-06
refactor: allow sound even iphone is unmuted

Levils114:

## Description
- Game audio now works even iPhone user is muted;
- The iOS layer was using the “ambient/solo ambient” session path, which is exactly the path that the mute switch can silence. I switched that to a playback session that stays audible under the user’s app volume settings.

## Related Issues
N/A


slime73:
Merged, thanks!

